PR #4, opened 2024-06-28
Help the user debugging stores.py

MeraX:
The code crashed with `DEBUG_ZARR_LOADING=1` and it took me a while to figure out what was going on until I realized, that the debugging assumes `DirectoryStore` while other formats are supported by `zarr.convenience.open` in principle.

### DIFF
--- a/src/anemoi/datasets/data/stores.py
+++ b/src/anemoi/datasets/data/stores.py
@@ -111,6 +111,9 @@ def open_zarr(path, dont_fail=False, cache=None):
 
         if DEBUG_ZARR_LOADING:
             if isinstance(store, str):
+                import os
+                if not os.path.isdir(store):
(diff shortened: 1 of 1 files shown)


src/anemoi/datasets/data/stores.py
```diff
@@ -111,6 +111,9 @@ def open_zarr(path, dont_fail=False, cache=None):
 
         if DEBUG_ZARR_LOADING:
             if isinstance(store, str):
+                import os
+                if not os.path.isdir(store):
+                    raise NotImplementedError("DEBUG_ZARR_LOADING is only implemented for DirectoryStore. Please disable it for other backends.")
```
MeraX, line 116: Alternatively, one could do
```suggestion
            import os
            if isinstance(store, str) and os.path.isdir(store):
```

Issue #46, opened 2024-09-18
As data is in the midst of conversion, getting "ERROR Error in retrieving metadata (cannot build data request info) for XArrayMetadata( ...)"


CSyl:
### What happened?

   With the latest release of anemoi-dataset, I was able to convert the zarr to an anemoi dataset, however several error messages regarding the metadata retrieval were showing up during the data creation/conversion process of the zarr data. The resulting error would state "ERROR Error in retrieving metadata (cannot build data request info) for XArrayMetadata( ...)." 

As a side note, I am able to compute the mean over the grid points of the resulting anemoi datasets, but am concerned about the errors I am getting. Is this an expected error? Could you please assist on why this error may be occurring?

Thank You!

### What are the steps to reproduce the bug?

The step-by-step of one of the case scenarios performed for which produces this error:

**Data Tested:**
A subset of ERA5 ZARR from GCP storage saved to local disk was performed via a .py script:

```
import xarray as xr
import gcsfs

gcp_ar_era5_subset = xr.open_zarr('gs://gcp-public-data-arco-era5/ar/1959-2022-1h-360x181_equiangular_with_poles_conservative.zarr', chunks={'time': 48}, consolidated=True)

gcp_ar_era5_subset = gcp_ar_era5_subset.isel(time=slice(-15, -1))

gcp_ar_era5_subset.to_zarr('gcp_ar_era5_subset.zarr')
```

**Configuration File Used:**
A yaml was created with the following content & saved as test_local_gcp_ar_sample_zarr.yaml

```
dates:
  start: 2021-12-31T09:00:00
  end: 2021-12-31T22:00:00
  frequency: 1h

input:
  xarray-zarr:
    url: "gcp_ar_era5_subset.zarr"
    param: [2m_temperature,
    10m_u_component_of_wind,
    geopotential,
    10m_v_component_of_wind,
    surface_pressure]
```

**Command Executed:**
>>> `anemoi-datasets create test_local_gcp_ar_sample_zarr.yaml test_local_gcp_ar_sample_zarr.zarr`

* Note: The latest release (v0.4.5) of **_anemoi-datasets_** released this past Friday (11/13/24) was used in this execution.

**Findings/Results:** 
[...]

PR #49, opened 2024-09-21
Bug fix when creating dataset from zarr

b8raoult:
Addressing https://github.com/ecmwf/anemoi-datasets/issues/46

### DIFF
--- a/src/anemoi/datasets/create/__init__.py
+++ b/src/anemoi/datasets/create/__init__.py
@@ -487,6 +487,7 @@ class Init(Actor, HasRegistryMixin, HasStatisticTempMixin, HasElementForDataMixi
         assert chunks == self.dataset.get_zarr_chunks(), (chunks, self.dataset.get_zarr_chunks())
 
         def sanity_check_config(a, b):
+            return
             a = json.dumps(a, sort_keys=True, default=str)
             b = json.dumps(b, sort_keys=True, default=str)
             b = b.replace("T", " ")  # dates are expected to be different because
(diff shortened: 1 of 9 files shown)


src/anemoi/datasets/create/__init__.py
```diff
@@ -487,6 +487,7 @@ def _run(self):
         assert chunks == self.dataset.get_zarr_chunks(), (chunks, self.dataset.get_zarr_chunks())
 
         def sanity_check_config(a, b):
+            return
```
mchantry, line 490: Why return here? Rest of code won't be used.
Should this be on line 498?

src/anemoi/datasets/create/input.py
```diff
@@ -108,14 +108,15 @@ def _data_request(data):
     for field in data:
         try:
             if date is None:
-                date = field.datetime()["valid_time"]
+                date = field.metadata("valid_datetime")
 
-            if field.datetime()["valid_time"] != date:
+            if field.metadata("valid_datetime") != date:
```
mchantry, line 113: What is this line doing?

mchantry: i.e. why continue?

src/anemoi/datasets/create/__init__.py
b8raoult: The message below has confused a few users. So, I have temporarily disabled it.

src/anemoi/datasets/create/input.py
b8raoult: We only care about one date. So we consider only the fields with the same valid date as the first.

floriankrb (approved):

PR #59, opened 2024-09-27
publish dataset

floriankrb:
Add anemoi-datasets publish, relying on anemoi-registry.

Save user recipe in the datasets.


floriankrb:
Best practice is to avoid writing usernames and passwords in the yaml recipe files.
But for extra security, all strings that could potentially contain passwords are filtered and removed from the final dataset.

b8raoult (dismissed):

PR #99, opened 2024-10-24
Feature/new datasets

mariahpope:
These are some helpful changes and/or additions that were necessary as I tested out the anemoi-datasets module on a few different datasets. Please let me know if you have any questions! Thanks!

_**Bug Fix**_

1. src/anemoi/datasets/create/functions/sources/xarray/variable.py
Bug fix - the arg in the class Variable is "variable" not "var". This caused a failure for me when working with levels until fixed.

_**Additions I found helpful**_

2. src/anemoi/datasets/create/__init__.py
Add the ability to work with cftime when needed.

3. src/anemoi/datasets/create/functions/sources/xarray/fieldlist.py
This is an update that was helpful when working with a curvilinear grid (but will not always be necessary - it depends on the dataset). It came after I noticed that some dimensions are not always set as coordinates/variables in the dataset. For example. a dataset with a curvilinear grid could have lat/lon as coordinates, but the relevant x and y index dimensions are not set as coordinates or variables. Therefore, this is a check to make sure that if dimension in a dataset was identified in a "flavour" that it is also set as a coordinate within the dataset.

_**Ability to read from Microsoft's Planetary Computer**_

There are a number of super useful datasets on planetary computer, and these updates make it possible to read most zarr stores from there (e.g. CONUS404 example: https://planetarycomputer.microsoft.com/dataset/conus404#Example-Notebook). The URL that a user of anemoi-datasets would use in a yaml file is the STAC collection URL associated with the dataset. I chose to set it up this way because this url appeared to be the most consistent way to reference zarr datasets within planetary computer.

4. src/anemoi/datasets/create/functions/sources/xarray/__init__.py
Elif statement specifically for planetary computer. There are a few extra hoops to jump through for open_zarr to work.

5. src/anemoi/datasets/data/stores.py
Add class PlanetaryComputerStore() that mimics the already existing S3 store and HTTPS store. 

**_Unit test addition_**

6. tests/xarray/test_zarr.py
I added two unit tests for additional datasets. The updates within this PR were able to make anemoi-datasets work with both of these datasets.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,6 +97,9 @@ optional-dependencies.tests = [
 optional-dependencies.xarray = [
   "gcsfs",
   "kerchunk",
+  "pandas", 
+  "pystac_client",
+  "planetary_computer",
 ]
 
 urls.Documentation = "https://anemoi-datasets.readthedocs.io/"
--- a/src/anemoi/datasets/create/__init__.py
+++ b/src/anemoi/datasets/create/__init__.py
@@ -7,6 +7,7 @@
 # nor does it submit to any jurisdiction.
 #
 
+import cftime
 import datetime
 import json
 import logging
@@ -66,6 +67,19 @@ def json_tidy(o):
 
     if isinstance(o, datetime.timedelta):
         return frequency_to_string(o)
+
+    if isinstance(o, cftime.DatetimeJulian):
+        import pandas as pd
+        
+        o = pd.Timestamp(
+            o.year,
+            o.month,
+            o.day,
+            o.hour,
+            o.minute,
+            o.second,
+        )
+        return o.isoformat()
 
     raise TypeError(repr(o) + " is not JSON serializable")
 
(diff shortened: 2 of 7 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-datasets?pullRequest=99) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-datasets?pullRequest=99) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-datasets?pullRequest=99) it.</sub>

b8raoult (commented):
Can you add the new dependencies in the `pyproject.toml` file, as optional dependecies? Namely pandas, pystac_client and planetary_computer.

mchantry:
Thanks for the great contribution!

b8raoult (approved):

PR #118, opened 2024-11-08
feat: Allow noncontinuous ranges of time to be given

HCookie:
- To be used in `anemoi.training`

Allows for a dataset to consist of seperate slices of time

----
📚 Documentation preview 📚: https://anemoi-datasets--118.org.readthedocs.build/en/118/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ Keep it human-readable, your future self will thank you!
 ### Changed
 - make test optional when adls is not installed Pull request #110
 
+### Added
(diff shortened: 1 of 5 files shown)


HCookie:
@floriankrb 
Do you have suggestion on how to implement this? 
- Evaluating all indices for a union between the date subset and the frequency?
- Concat all slices of datasets?

b8raoult (changes requested):
This does not work because the `frequency` changes at each boundary. If someone use `ds[i] and ds[i+1]` they expect both samples to be 1 `frequency` apart. They can be many years apart at the boundary between two ranges.

b8raoult:
We close it for now. We can always re-open it.

PR #144, opened 2024-11-21
Added reorder sort string, which sorts variables alphabetically.

einrone:
**Quick summary:**
Added another reordering option reorder="sort" which sorts the variables alphabetically to enable transfer learning for stretched grid.

**Why, and what is the purpose?**

During transfer learning of stretched-grid model, the cutout and the option adjust='all' functionality is used. This will result that the combined dataset (ds1 and ds2) will have their variable list sorted alphabetically. However, then the pretrained model 
needs to follow this type of sorting. This implementation will reorder the variable list to sort them alphabetically, hence enabling transfer learning when use multiple datasets when using reorder='sort'.

I have done some test, and also created some units tests, but unsure on where to put LAM dataset in order to make the testing work. For now I will submit this as a draft PR, if unit test is needed I will also include them.

### DIFF
--- a/src/anemoi/datasets/data/dataset.py
+++ b/src/anemoi/datasets/data/dataset.py
@@ -249,13 +249,19 @@ class Dataset:
         return sorted([v for k, v in self.name_to_index.items() if k not in vars])
 
     def _reorder_to_columns(self, vars):
+        if isinstance(vars, str) and vars=="sort":
+            # Sorting the variables alphabetically.
+            # This is cruical for pre-training then transfer learning in combination with
+            # cutout and adjust = 'all'
+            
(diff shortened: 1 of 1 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-datasets?pullRequest=144) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-datasets?pullRequest=144) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-datasets?pullRequest=144) it.</sub>

Issue #155, opened 2024-12-03
Problems building a 3D hypercube when selecting multiple variables


mariahpope:
**Describe the bug**
I am trying to build a dataset and am running into a bug while the module tries to build a 3D hypercube. This only happens when there is more than one variable.

** Version number **
The most current version of the develop branch and pyproject.toml (as of Dec 3rd, 2024)

**To Reproduce**
1. Used a yaml with this configuration:
```
dates:
  start: 2021-01-01T00:00:00
  end: 2021-01-01T23:00:00
  frequency: 6h

input:
    xarray-zarr:
        url: "gs://gcp-public-data-arco-era5/ar/1959-2022-6h-128x64_equiangular_conservative.zarr"
        param:
            - 2m_temperature
            - surface_pressure
```
2. Ran:
`anemoi-datasets create era5_test.yaml era5_test.zarr`

**URL to sample input data**
gs://gcp-public-data-arco-era5/ar/1959-2022-6h-128x64_equiangular_conservative.zarr

**Expected behavior**
This should have created a dataset with both 2m_temperature and surface_pressure, but it results in a hypercube error (see screenshot below). If I switch to only one variable (e.g. just 2m_temperature) it works just fine.

**Screenshots**
<img width="1064" alt="Screenshot 2024-12-03 at 1 45 55 PM" src="https://github.com/user-attachments/assets/1221dc68-be57-4912-9387-fe920349deab">

b8raoult:
Can you try https://github.com/ecmwf/anemoi-datasets/pull/159 ?

mariahpope:
Yes, just tried it! It seems to work now for the example I initially provided above (the era5 zarr with temperature and pressure)!
However, I did a few more tests out of curiosity and found that it will provide the same error if I try to add variables with additional levels. For example, if I use this yaml (same data, just add temperature levels 50 and 100):
```
dates:
  start: 2021-01-01T00:00:00
  end: 2021-01-01T23:00:00
  frequency: 6h

input:
  join:
  - xarray-zarr:
        url: "gs://gcp-public-data-arco-era5/ar/1959-2022-6h-128x64_equiangular_conservative.zarr"
        param:
            - 2m_temperature
            - surface_pressure
  - xarray-zarr:
        url: "gs://gcp-public-data-arco-era5/ar/1959-2022-6h-128x64_equiangular_conservative.zarr"
        param:
            - temperature
        level:
            - 50
            - 100
```
I get this similar error:
[...]

Issue #116, opened 2024-11-07
speed improvement


mariahpope:
As I work with very large zarr stores in the cloud (such as an ERA5 dataset that goes back 60+ years) I have noticed that the framework gets hung up and takes a very long time to run even if I only want to build a small dataset (e.g. a couple days or a month for a test) from it. After some digging I noticed that the current setup loops through all time steps in the dataset as it loads. This means that if I only want a day or two of data from a zarr store that is 60+ years it will loop through the whole store no matter what. Is this behavior intended/expected? Or could there be a fix to only look at data between dates that were specified in the yaml?

A current idea would be to update lines 58-63 in [__init__.py](https://github.com/ecmwf/anemoi-datasets/blob/develop/src/anemoi/datasets/create/functions/sources/xarray/__init__.py) to subset the store to a time range of interest:

Current code:
```
    fs = XarrayFieldList.from_xarray(data, flavour)

    if len(dates) == 0:
        return fs.sel(**kwargs)
    else:
        result = MultiFieldList([fs.sel(valid_datetime=date, **kwargs) for date in dates])
```

Possible update/solution could look something like this:
```
    if len(dates) == 0:
        fs = XarrayFieldList.from_xarray(data, flavour)
        return fs.sel(**kwargs)
    elif len(dates) == 1:
        data = data.sel(time=dates[0])
        fs = XarrayFieldList.from_xarray(data, flavour)
        result = MultiFieldList([fs.sel(valid_datetime=date, **kwargs) for date in dates])
    else:
        data = data.sel(time=slice(dates[0], dates[-1]))
        fs = XarrayFieldList.from_xarray(data, flavour)
        result = MultiFieldList([fs.sel(valid_datetime=date, **kwargs) for date in dates])  
```
This would clip a store to just the first time step that is needed to initialize the dataset, and then would clip to the time range specified in the yaml before starting the loading process. I tested this out and it seems to work, and it greatly speeds up the framework when working with subsets of large zarr stores. Is there a reason that the code needs to process the whole store? If this is an update that would be helpful I can gladly start a PR! Would love to hear your thoughts. 

Thank you!!

floriankrb:
Thanks for the idea, improving build speed is also important for `anemoi-datasets`, and a PR that makes it more efficient would be indeed welcome.

[...]

PR #159, opened 2024-12-13
Fix for #155 and #116

b8raoult:
A fix for https://github.com/ecmwf/anemoi-datasets/issues/116 and https://github.com/ecmwf/anemoi-datasets/issues/155


timothyas:
Thank you @b8raoult!

PR #162, opened 2024-12-19
feat: support list indices for sampling

JPXKQX:
This PR addresses that the following command fails when `grid_shard_indices` is a list of indices.

```
x = self.data[100:105, :, :, grid_shard_indices]
```


b8raoult (commented):
I am not sure that will work for all combinations of zarr (join, concat, etc)

aaron-hopkinson:
Closing after discussing with @b8raoult as this PR is no longer relevant. I've opened #406 in an attempt to capture the problem

PR #220, opened 2025-02-26
feat: zarr3

floriankrb:
## Type of Change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

## Code Compatibility

-   [ ] I have performed a self-review of my code

### Code Performance and Testing

-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I ran the [complete Pytest test](https://anemoi.readthedocs.io/projects/training/en/latest/dev/testing.html) suite locally, and they pass

### Dependencies

-   [ ] I have ensured that the code is still pip-installable after the changes and runs
-   [ ] I have tested that new dependencies themselves are pip-installable.

### Documentation

-   [ ] My code follows the style guidelines of this project
-   [ ] I have updated the documentation and docstrings to reflect the changes
-   [ ] I have added comments to my code, particularly in hard-to-understand areas


anaprietonem:
@floriankrb was the conclusion in terms of adding updating to zarr 3?

floriankrb:
@anaprietonem I updated the description of the PR

frazane:
Just adding a small comment here. You've probably considered this already but here we go. 

The transition to Zarr v3 specification comes with a nice opportunity: we could use the [sharding](https://zarr.readthedocs.io/en/latest/user-guide/performance.html#sharding) feature! It would allow us to have chunked variables (avoiding having to load them all in memory before taking a subset) while still retaining very high read speeds. Basically we would make each timestamp a shard, and then chunk variables inside the shard. The downside is slower write speed, but it’s not very important.

b8raoult:
All tests are now passing with zarr2 and zarr3. At the moment of writing, zarr3 still does not support datetime64 (https://github.com/zarr-developers/zarr-python/issues/2616)

Tests are much more slower with zarr3, this will require more investigation. For example `test_slice_4` takes 98s with zarr2 (1.5 minutes) and 612s with zarr3 (10 minutes)

The profiler shows for zarr2:

```
   773147    0.339    0.000   95.489    0.000 .../anemoi-datasets/src/anemoi/datasets/data/stores.py:148(__getitem__)
   773165    1.553    0.000   95.152    0.000 .../python3.12/site-packages/zarr/core.py:656(__getitem__)
```
   
and for zarr3:

```
   773147    0.449    0.000  615.551    0.001 .../anemoi-datasets/src/anemoi/datasets/data/stores.py:148(__getitem__)
   773165    2.088    0.000  615.118    0.001 .../python3.12/site-packages/zarr/core/array.py:2294(__getitem__)
```

The test uses an in-memory zarr store.

src/anemoi/datasets/create/__init__.py
```diff
@@ -609,6 +615,32 @@ def __init__(
...
+            pytesting = "PYTEST_CURRENT_TEST" in os.environ
+
+            if pytesting or force_zarr3:
+                LOG.warning("⚠️" * 80)
+                LOG.warning("Zarr version 3 is used, but this is an unsupported feature.")
+                LOG.warning("⚠️" * 80)
+            else:
+                LOG.warning("⚠️" * 80)
+                LOG.warning(
+                    f"Only Zarr version 2 is supported when creating datasets, found version: {zarr.__version__}"
```
tjhunter, line 635: we should not support writing in zarr3 format for the time being, only reading. Is there a use case for it?

anaprietonem:
Hey @TomNicholas, thanks for the interest! Indeed, we’re aware of the growing interest in Zarr v3, and that’s what this PR is aiming to address.

Where we are at the moment is that some changes are needed to enable Zarr v3 support in anemoi-datasets—those included in this PR, but also things like dropping support for Python 3.10. We already have many datasets built with Zarr v2, and it’s important for us to benchmark runtime performance for training workloads. We’ve done some analysis here: https://github.com/ecmwf/anemoi-datasets/issues/486,  feel free to take a look and share your thoughts. Our tests on HPC systems show some degradation in runtime performance when moving to the Zarr v3 Python package.

We’re still working on gathering more data points and exploring ways to improve this. This was discussed with @ecmwf/anemoi_technical_subgroup, and the decision was to merge Zarr v3 support into main in anemoi-datasets, but not yet remove the pyproject constraint on Zarr v2. This way, users who want to experiment with Zarr v3 can easily force it, while anemoi-datasets officially supports it. If we find ways to further optimise performance, we can simply update the pyproject constraints in the future.

@floriankrb @b8raoult @cathalobrien  feel free to add to this!

b8raoult:
To complement @anaprietonem's message. We have a very thin compatibility wrapper around zarr2 and zarr3 in our code, for apis that have changed. This wrapper will use whatever version of zarr is installed. By default we mention version 2 in the dependencies, but a user can force a pip install of version 3, and the code will still work, with the caveat that zarr2 cannot read zarr3 datasets. Once we fully adopt zarr3, we will remove that wrapper and call the zarr api directly.

TomNicholas:
I don't think anyone has worked on that yet, no. 

But is that a blocker for you? Your response [here](https://github.com/zarr-developers/zarr-python/issues/3524#issuecomment-3414619162)

> You may close this issue if you wish.

implied that it is not a blocker. But if it is a blocker then we would like to help unblock you.

PR #304, opened 2025-04-29
test: add test for dataset creation using GRIB files with unstructured grid

frazane:
## Description

Adds tests to test the `grid_definition` argument of the `grib` source, which allows specifying grid definition files required when GRIB data is on an unstructured grid (e.g. for the ICON model). Tests are added for both MeteoSwiss data and DWD data (@MeraX).

## Summary of changes
- added data (sample grib files) to the test data server
- added a test for the source


MeraX (approved):
LGTM

mchantry (approved):
LGTM, thanks for your contribution.

PR #362, opened 2025-06-23
fix: remove trailing slash from zarr paths. 

matschreiner:
## Description

This PR adds a small fix to the Anemoi datasets interface that removes a trailing slash from .zarr/ paths passed to open_dataset. Without this fix, providing a path like "path/to/dataset.zarr/" results in a misleading ValueError, even when the dataset exists.

With this change, both
```
open_dataset("path/to/dataset.zarr/")
```
and
```
open_dataset("path/to/dataset.zarr")
```

will correctly resolve and open the dataset and log a warning that the path has been changed .

This is a bugfix. The open_dataset function fails to locate .zarr datasets when the input path ends with a trailing slash.


floriankrb (approved):

mchantry:
Thanks for the contribution.

PR #381, opened 2025-07-16
feat: Extend xarray Variable.sel to handle scalar variables

lbdreyer:
## Description
This modifies the code for `anemoi.datasets.create.sources.xarray_support.variable.Variable.sel()`to handle the situation where a `Variable` is scalar. This also includes some basic unit testing.

When trying to create datasets from netcdf files, I would hit an issue when one of my coordinates was a scalar as [this line would fail](https://github.com/ecmwf/anemoi-datasets/blob/2baa24ec5e596221fd06bcc24c0f37c85f3fdeff/src/anemoi/datasets/create/sources/xarray_support/variable.py#L224).

I am working with lots of netcdf files where each netcdf file represents a single time step so the time coordinate is scalar

n/a

##  Additional notes ##

### DIFF
--- a/src/anemoi/datasets/create/sources/xarray_support/variable.py
+++ b/src/anemoi/datasets/create/sources/xarray_support/variable.py
@@ -221,19 +221,22 @@ class Variable:
                 LOG.warning(f"Could not find {k}={v} in {c}")
             return None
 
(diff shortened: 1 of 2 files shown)


aaron-hopkinson (approved):
Thanks for this. 😄

PR #377, opened 2025-07-08
fix: define the missing forwards_subclass_metadata_specific method in the ZipBase class

eliott-lumet:
## Description

This PR defines the `forwards_subclass_metadata_specific` method in the  `ZipBase `class. This definition seems mandatory as `ZipBase` inherits from the `Forward`s class which declares an abstract method `forwards_subclass_metadata_specific`, see: [https://docs.python.org/3/library/abc.html#abc.abstractmethod](https://docs.python.org/3/library/abc.html#abc.abstractmethod)

## What problem does this change solve?

This change solves the following issue:

```
File "/perm/fra6527/venvs/3-aromeds-dev-7e65c1cb35/lib/python3.11/site-packages/anemoi/training/data/datamodule.py", line 98, in data_indices
    return IndexCollection(self.config, self.ds_train.name_to_index)
                                        ^^^^^^^^^^^^^
  File "/usr/local/apps/python3/3.11.10-01/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/perm/fra6527/venvs/3-aromeds-dev-7e65c1cb35/lib/python3.11/site-packages/anemoi/training/data/datamodule.py", line 103, in ds_train
    open_dataset(
  File "/perm/fra6527/venvs/3-aromeds-dev-7e65c1cb35/lib/python3.11/site-packages/anemoi/datasets/data/__init__.py", line 89, in open_dataset
    ds = _open_dataset(*args, **kwargs)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/perm/fra6527/venvs/3-aromeds-dev-7e65c1cb35/lib/python3.11/site-packages/anemoi/datasets/data/misc.py", line 506, in _open_dataset
    sets.append(_open(a))
                ^^^^^^^^
  File "/perm/fra6527/venvs/3-aromeds-dev-7e65c1cb35/lib/python3.11/site-packages/anemoi/datasets/data/misc.py", line 384, in _open
    return _open_dataset(**a).mutate()
           ^^^^^^^^^^^^^^^^^^
  File "/perm/fra6527/venvs/3-aromeds-dev-7e65c1cb35/lib/python3.11/site-packages/anemoi/datasets/data/misc.py", line 591, in _open_dataset
    sets.append(_open(a))
                ^^^^^^^^
[...]

PR #388, opened 2025-07-24
Fix: implement missing abstract method in Zip

paulina-t:
## Description
This pull request addresses an issue where the `Zip` class could not be instantiated because the abstract method `forwards_subclass_metadata_specific` was not implemented. The `Zip` class inherits this abstract method from the `Forwards` class, and its absence caused a `TypeError` when attempting to use the `Zip` class.

This change resolves the `TypeError: Can't instantiate abstract class Zip with abstract method forwards_subclass_metadata_specific` by implementing the `forwards_subclass_metadata_specific` method in the `Zip` class. For now, the method currently returns an empty dictionary as a placeholder.

The issue was encountered when calling `open_dataset(zip=[dataset_1, dataset_2])`.

### DIFF
--- a/src/anemoi/datasets/data/xy.py
+++ b/src/anemoi/datasets/data/xy.py
@@ -213,7 +213,17 @@ class ZipBase(Combined):
 
 class Zip(ZipBase):
     """Class for handling zipped datasets."""
+    def forwards_subclass_metadata_specific(self) -> Dict[str, Any]:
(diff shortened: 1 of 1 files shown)


eliott-lumet:
Hi Paulina, I've encountered and already reported the same issue: #377 
It should be "merged ASAP".

PR #408, opened 2025-08-28
fix: Added TestingContext for plugins

NatalieZelenka:
## Description
Fixes https://github.com/ecmwf/anemoi-plugins/issues/22

Bugfix for `anemoi-plugins` testing.

[`anemoi-plugins/issues/22`]( https://github.com/ecmwf/anemoi-plugins/issues/22)

### DIFF
--- a/src/anemoi/datasets/create/testing.py
+++ b/src/anemoi/datasets/create/testing.py
@@ -1,4 +1,4 @@
-
 class TestingContext:
     """A context for testing plugins."""
+
     pass


aaron-hopkinson:
@HCookie – is it worth changing what plugins expects?

HCookie:
@aaron-hopkinson The context is a good idea, but badly used atm, and at times does nothing. As we improve tests, this becomes viable. For now, I guess we can remove it from anemoi-plugins

HCookie (approved):

PR #430, opened 2025-10-03
refactor: reduce complexity of legacy sources

aaron-hopkinson:
## Description
Reduces the complexity of legacy sources by moving to straightforward class inheritance and implementation, removing the need for the decorator and dynamically generated classes. 


aaron-hopkinson:
NB: it may be helpful to review this change commit-by-commit – each of the commits is relatively small, and the tests should pass at each stage. The second commit looks like it is introducing a large change, but it's just moving functions from one part of a file to another (making what was previously a standalone function into a staticmethod on each of the classes) – I suspect some automatic formatting caused by the pre-commit hooks is confusing git and making the diff look bigger than it is.

b8raoult (approved):

PR #452, opened 2025-10-21
fix: tendencies source context, metadata keys, dates

### LINKED ISSUE #451: Tendencies source has wrong interface
During #430 and #450, I noticed that the `tendencies` source has the wrong interface, i.e. all other sources accept a "context" as a first argument, whereas this source doesn't. This will also be a problem when it calls the `mars` source under the hood, which is also expecting a context but does not receive one.

We need to investigate whether or not this causes an error (most likely) and fix the bug

aaron-hopkinson:
## Description

Closes #451 

During https://github.com/ecmwf/anemoi-datasets/pull/430 and https://github.com/ecmwf/anemoi-datasets/pull/450, I noticed that the tendencies source has the wrong interface, i.e. all other sources accept a "context" as a first argument, whereas this source doesn't. This will also be a problem when it calls the mars source under the hood, which is also expecting a context but does not receive one. This PR adds the missing context, which is unused by the tendencies source directly, but is passed through directly to the mars source.

Additionally, I had to cast `dates` to a `list` as it was a `GroupOfDates`, and change some of the metadata checks due to updates in earthkit-data.

### DIFF
--- a/src/anemoi/datasets/create/sources/tendencies.py
+++ b/src/anemoi/datasets/create/sources/tendencies.py
@@ -16,6 +16,7 @@ from earthkit.data.readers.grib.output import new_grib_output
 
 from anemoi.datasets.create.sources import source_registry
 
+from ...dates.groups import GroupOfDates
 from .legacy import LegacySource
 
 
@@ -87,15 +88,15 @@ def group_by_field(ds: Any) -> dict[tuple, list[Any]]:
 class TendenciesSource(LegacySource):
 
     @staticmethod
-    def _execute(context: Any, dates: list[datetime.datetime], time_increment: Any, **kwargs: Any) -> Any:
+    def _execute(context: Any, dates: GroupOfDates, time_increment: Any, **kwargs: Any) -> Any:
         """Computes tendencies for the given dates and time increment.
 
         Parameters
         ----------
         context : Any
             The source context - passed through directly to the mars source.
-        dates : List[datetime.datetime]
-            A list of datetime objects.
+        dates : GroupOfDates
+            An object representing the dates for which to compute tendencies.
         time_increment : Any
             A time increment string ending with 'h' or a datetime.timedelta object.
         **kwargs : Any


src/anemoi/datasets/create/sources/tendencies.py
```diff
@@ -139,12 +141,14 @@ def _execute(dates: list[datetime.datetime], time_increment: Any, **kwargs: Any)
             print("❌", k)
 
             for field, b_field in zip(group1[k], group2[k]):
-                for k in ["param", "level", "number", "grid", "shape"]:
+                for k in ["param", "level", "number"]:
```
aaron-hopkinson: (NB: shape here was redundant because it is checked a few lines below)

src/anemoi/datasets/create/sources/tendencies.py
```diff
@@ -139,12 +141,14 @@ def _execute(dates: list[datetime.datetime], time_increment: Any, **kwargs: Any)
...
                     assert field.metadata(k) == b_field.metadata(k), (
                         k,
                         field.metadata(k),
                         b_field.metadata(k),
                     )
+                # should we check grids are the same? e.g.
+                # assert field.metadata().gridspec == b_field.metadata().gridspec
```
aaron-hopkinson: This check won't actually work - so it's a bit "unsafe" at the moment, but does at least run (unlike before)...

Since the shapes still have to be identical, we might catch some significant grid mismatches, but not ones where we have the same number of points but different e.g. areas

src/anemoi/datasets/create/sources/tendencies.py
anaprietonem: and what about grid? - never mind reading comment below

anaprietonem (approved):
The changes looks good to me! Only aspect I think could be worth clarifying is the grid. @b8raoult ?

src/anemoi/datasets/create/sources/tendencies.py
aaron-hopkinson: `gridspec` is a fairly complex object (as far as I know), and I think still a WIP in ekd. Doing `==` will fail here, even when I know the underlying grid is the same. Might be a bug in ekd, might be my misunderstanding. 

(Another alternative is e.g. to compare `field.metadata('gridType')`, but that will only compare if e.g. both are `"regular_ll"`, rather than something containing more info (like the gridspec) which presumably contains info about area, resolution etc)..

aaron-hopkinson:
@anaprietonem 

> these tendencies source is the one used to compute the statistics right?

I don't think so. I think the statistics that are stored in the dataset are calculated elsewhere in the code. If I had to guess, this is a source that was used in earlier versions of anemoi-datasets, but hasn't been used for a while... (which is why it's missed out on being documented, tested etc).

> by default we compute statistics using the frequency of the dataset?

Pass (without looking through the code).


> If a user uses this source then it essentially overwrites the default tendencies or would it add a second set of statistics

I don't think it writes different statistics, but instead produces new "fields" which represent something that looks like a difference between fields – i.e. the `2t` field isn't 2m temperature, but instead a difference of 2m temps...

> could use with an example recipe and a paragraph to summarise how to use this source and what's the behaviour

I suspect someone else is far better qualified than me to do that 😆

Issue #456, opened 2025-10-29
MatchingFieldsFilter not working for xarray-zarr sources


SimonKamuk:
MatchingFieldsFilters do not seem to be working for xarray-zarr sources.
Huge thanks to @mpvginde for identifying the two issues:
- XArrayField.to_numpy() returns xarray DataArray instead of numpy array
- XArrayMetadata contains both 'variable' and 'param' so grouping by the metadata doesn't work because GroupByParam._get_groups only takes param into account by popping it and looking at the remaining metadata, which then doesn't match as the variable metadata still doesn't match.

PR #457, opened 2025-10-29
fix: xarray-zarr filters

SimonKamuk:
## Description
The PR solves an issue where filters didn't work with xarray-zarr sources

This is a bugfix for two issues:
- In `XArrayField.to_numpy()` with `flatten=True`, the output is `values.values.flatten()` but for `flatten=False` the output is only `values` where it should be `values.values` to actually get the numpy array.
- XArrayMetadata contains both 'variable' and 'param' so grouping by the metadata doesn't work because GroupByParam._get_groups only takes param into account by popping it and looking at the remaining metadata. If the remaining metadata matches then the params are grouped together. This doesn't work when variable is also there, because now the remaining metadata doesn't match

https://github.com/ecmwf/anemoi-datasets/issues/456

##  Additional notes ##

### DIFF
--- a/src/anemoi/datasets/create/sources/xarray_support/field.py
+++ b/src/anemoi/datasets/create/sources/xarray_support/field.py
@@ -121,16 +121,16 @@ class XArrayField(Field):
             Index to select a specific element, by default None.
         """
         if index is not None:
-            values = self.selection[index]
+            values = self.selection[index].values
         else:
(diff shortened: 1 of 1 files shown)


aaron-hopkinson:
(Checked with @mpvginde – he's happy for this to be merged, so merging). Thanks again @SimonKamuk

PR #462, opened 2025-11-03
chore: remove tendencies source

### LINKED ISSUE #451: Tendencies source has wrong interface

aaron-hopkinson:
## Description
Removes unused `tendencies` source.

Also closes #451.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,10 @@
 # anemoi-datasets
 
-**DISCLAIMER**
-This project is **BETA** and will be **Experimental** for the foreseeable future.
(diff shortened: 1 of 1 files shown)


anaprietonem (approved):
good to be merged

PR #480, opened 2025-11-19
feat: Observations feature branch

mchantry:
Planned medium-term feature branch to enable the development of observation dataset training.

Key components:
- [x] Call structure/inference.
- [ ] First sources.
- [ ] Dataset creation from sources.
- [x] Dataset loading.

Screenshot 2025-11-19 at 13 23 56
Screenshot 2025-11-19 at 13 34 46

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

----
📚 Documentation preview 📚: https://anemoi-datasets--480.org.readthedocs.build/en/480/

PR #491, opened 2025-11-30
feat: first obs recipe

b8raoult:
## Description
To be merged into #480, this branch develops recipes for observation dataset creation.
The description of the changes are being documented [here](https://github.com/ecmwf/anemoi-datasets/blob/feat/first-obs-recipe/docs/changes.md).
The ADR describing the design choices can be seen [here](https://github.com/ecmwf/anemoi-datasets/blob/feat/first-obs-recipe/docs/adr/adr-1.md).

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-datasets--491.org.readthedocs.build/en/491/

### DIFF
--- a/tests/create/test_sources.py
+++ b/tests/create/test_sources.py
@@ -8,6 +8,7 @@
 # nor does it submit to any jurisdiction.
 
 import os
+import shutil
 import sys
 
 import numpy as np
@@ -511,6 +512,8 @@ def test_csv(get_test_data: callable) -> None:
 
 # @pytest.mark.skip(reason="ODB source currently not functional")
 @skip_if_offline
+@skip_if_offline
+@pytest.mark.skipif(shutil.which("odc") is None, reason="odc command not accessible")
 def test_odb(get_test_data: callable) -> None:
     from anemoi.datasets.create.sources import create_source
     from anemoi.datasets.dates.groups import GroupOfDates


src/anemoi/datasets/tabular/caching.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+class _Chunk:
+
+    def __init__(self, array: zarr.Array, chunk_index: int):
+        self.array = array
+        self.chunk_index = chunk_index
+        self.chunk_size = array.chunks[0]
+        self.offset = chunk_index * self.chunk_size
+        self.cache = array[self.offset : min(self.offset + self.chunk_size, array.shape[0])]
```
aaron-hopkinson:
```suggestion
        # use min to avoid indexing out of bounds of array
        self.cache = array[self.offset : min(self.offset + self.chunk_size, array.shape[0])]
```

src/anemoi/datasets/tabular/caching.py
```diff
@@ -0,0 +1,255 @@
...
+        self.dirty = False
+        self.loaded = False
```
aaron-hopkinson: Unused. Can we remove this?

src/anemoi/datasets/tabular/caching.py
```diff
@@ -0,0 +1,255 @@
...
+            self.array[self.offset : self.offset + len(self.cache)] = self.cache
+            self.dirty = False
+
+    def _remove(self, lru, why):
+        if LOG.isEnabledFor(logging.DEBUG):
+            LOG.debug(f"Removing chunk {self.chunk_index} {self} because {why}")
+        self.flush()
+        del lru[self.chunk_index]
+
+    def resize(self, lru, new_array_shape):
```
aaron-hopkinson: Do we need to support this?

src/anemoi/datasets/tabular/caching.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def __setitem__(self, key, value):
+        chunk, key = self._get_chunk(key)
+        chunk[key] = value
+
+    def __getitem__(self, key):
+        chunk, key = self._get_chunk(key)
+        return chunk[key]
+
+    def flush(self, reset=False):
```
aaron-hopkinson: `flush` doesnt remove from LRU cache unless `reset=True`.  This is for flushing changes, rather than clearing the cache – maybe we need a comment to make that clear (since that wasn't my initial impression of what it was used for).

What's the use case for flushing and resetting? (or flush but not reset?) Should this be two methods (with one calling the other)?

But probably not needed unless we need to support buffered writing?

src/anemoi/datasets/tabular/caching.py
```diff
@@ -0,0 +1,255 @@
...
+        for chunk in self._chunks.values():
+            chunk.flush()
+
+        if reset:
+            self._chunks.clear()
+
+    def _ensure_chunk(self, chunk_index):
```
aaron-hopkinson: I think it would be clearer if this method were renamed to `_get_chunk` (and the current `_get_chunk` renamed to something like `_get_chunk_key` or similar)

src/anemoi/datasets/tabular/caching.py
```diff
@@ -0,0 +1,255 @@
...
+                LOG.debug(
+                    f"Loaded chunk {chunk_index} {self._chunks[chunk_index]}. Cached: {sorted(self._chunks.keys())},"
+                    f" {sum(chunk.cache.nbytes for chunk in self._chunks.values()) / 1024 / 1024:.2f} MB"
+                    f" dirty={sum(chunk.dirty for chunk in self._chunks.values())}",
+                )
+        return self._chunks[chunk_index]
+
+    def _normalise_slice(self, key):
+        # Convert to absolute indices (no Nones, no negatives)
+        return slice(*key.indices(self._arr.shape[0]))
```
aaron-hopkinson: We use `self._arr.shape[0]` in a few different places. Is it worth saving this as an instance variable? It probably has some value as an explanatory variable

src/anemoi/datasets/tabular/caching.py
```diff
@@ -0,0 +1,255 @@
...
+    def _get_chunk(self, key):
+        """Retrieves the appropriate chunk and a normalised key."""
+        match key:
+
+            case int():
+                chunk_index = key // self._chunk_0
+                return self._ensure_chunk(chunk_index), key
+
+            case slice():
+                return self._get_chunk(key, slice(None, None, None))
```
aaron-hopkinson: I thin this will fail. `self._get_chunk` only takes one value

src/anemoi/datasets/tabular/caching.py
```diff
@@ -0,0 +1,255 @@
...
+
+            case tuple():
+                if isinstance(key[0], int):
+                    chunk_index = key[0] // self._chunk_0
+                    return self._ensure_chunk(chunk_index), key
+
+                if isinstance(key[0], slice):
+                    key = (self._normalise_slice(key[0]),) + key[1:]
```
aaron-hopkinson: The contents of this this branch feel like they belong in `case slice():` and then if we remove the need to return the key, we could just call this recursively?

src/anemoi/datasets/tabular/caching.py
```diff
@@ -0,0 +1,255 @@
...
+                    indices = set(a // self._chunk_0 for a in range(*key[0].indices(self._arr.shape[0])))
+                    if len(indices) == 1:
+                        chunk_index = indices.pop()
+                        return self._ensure_chunk(chunk_index), key
+                    else:
+                        return _MultiChunkSlice(self, key), key
```
aaron-hopkinson: Related to above, this bit a bit hard to follow without examples - it's not clear when `MultiChunkSlice` is triggered to me

aaron-hopkinson (changes requested):
I've only reviewed `src/anemoi/datasets/tabular/caching.py` – the code looks like it should work correctly aside from one potential bug (`case slice()` in `ChunksCache._get_chunk`). The rest of the comments either relate to making the code clearer to someone unfamiliar with the code, or relating to scope of the work (a friendly challenge 😃)

docs/layouts/tabular.rst
```diff
@@ -0,0 +1,331 @@
...
+
+   sample.time_deltas
+
+   assert len(sample.time_deltas) == number_of_observations_in_window
+
+Slices
+------
+
+When slicing the dataset, the same rules apply as for indexing with an
+integer, but you can recover the samples using the ``boudaries``
```
mchantry:
```suggestion
integer, but you can recover the samples using the ``boundaries``
```

PR #494, opened 2025-12-02
feat: Add xarray patches for changing time dimension and summing variables

SimonKamuk:
## Description
This PR adds xarray patches for changing time dimension and summing variables

This new feature is added to solve the issue where e.g. danra forecast zarr data cannot be read by anemoi datasets. This was because the time dimensions in danra forecasts are analysis time and lead time, but anemoi datasets needs the valid time.

I solved it by adding an xarray patch which converts the time dimensions.

I also added a patch which sums variables, which is needed for accumulated fields. In the danra zarr files, accumulated variables contain hourly values, so these need to be summed to match the time steps used for training.

Lastly, I added a patch for selecting in the dataset with xarrays sel

The patches could be applied like this:

```
patch:
  subset_dataset:
    forecast_duration:
    - "1:00:00"
    - "2:00:00"
    - "3:00:00"
  analysis_lead_to_valid_time:
    analysis_time_coordinate: analysis_time
    lead_time_coordinate: forecast_duration
    valid_time_coordinate: time
  rolling_sum:
    dim: time
    steps: 3
    vars:
    - tp_accum_past_hr
```


HCookie (changes requested):
Thank you for the contribution, I have a couple comments mainly on the typing.
If you could fix that, we can look at merging.

PR #496, opened 2025-12-10
feat(apply mask): apply a boolean mask when opening a dataset

clairemerker:
## Description
This PR introduces `mask` as an argument to `open_dataset`. This argument allows to apply an arbitrary spatial boolean mask from a .npy file to the dataset.

It is a new feature for `open_dataset` that allows to select arbitrary spatial points of a dataset, it could be used e.g. to thin data on an irregular grid or select a region of interest. 
It was tested to reduce the density of grid points of the triangular grid of the ICON model. The plots show a zoom over Switzerland with the dense grid at approx. 1km resolution and the grid after applying a mask selecting only every second triangle. 

Original grid:
image

Half the points masked using external mask:
image

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-datasets--496.org.readthedocs.build/en/496/


frazane:
> One thing to note is that this will not automatically carry over to inference.

@gmertes while I agree this is true when doing inference from files (e.g. GRIB), what about when just using `input: test` or similar? Shouldn't this work?

gmertes:
Downstream ci is fixed. For me this PR is good to go, but did you test that the behaviour in inference is as you expect?

Issue #498, opened 2025-12-17
Selecting a subset of levels while creating Zarr file from multiple level NetCDF files


mo-akm:
**Describe the bug**
Context:

Making Anemoi Zarr file from netcdf files.

The files contain data on 33 pressure levels in Pa, with levels as follows:
 1000, 2000, 3000, 4000, 5000, 7000, 10000, 12500, 15000, 17500, 20000, 22500, 25000, 27500, 30000, 32500, 35000, 37500, 40000, 45000, 50000, 55000, 60000, 65000, 70000, 75000, 80000, 85000, 90000, 92500, 95000, 97500, 100000

I wish to sub-select the levels with the following example entries in my recipe file:

```yaml
pressure: &pressure_levels
  - 3000
  - 7000
  - 10000
  - 15000
  - 20000
  - 25000
  - 30000
  - 40000
  - 50000
  - 70000
  - 85000
  - 95000
  - 100000

input:
  join:
    - netcdf:
      path: /path/to/files/*-relative_humidity_on_pressure_levels.nc
      param: relative_humidity
      pressure: *pressure_levels
```

Problem:
As soon as I include a level with a pressure value from 35000 upwards, I get the following error message

```bash
File ".pixi/envs/default/lib/python3.14/site-packages/anemoi/datasets/create/sources/xarray_support/fieldlist.py", line 228, in sel
    v = v.sel(missing, **rest)
  File ".pixi/envs/default/lib/python3.14/site-packages/anemoi/datasets/create/sources/xarray_support/variable.py", line 213, in sel
    i = c.index(v)
  File ".pixi/envs/default/lib/python3.14/site-packages/anemoi/datasets/create/sources/xarray_support/coordinates.py", line 168, in index
    return self._index_multiple(value)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File ".pixi/envs/default/lib/python3.14/site-packages/anemoi/datasets/create/sources/xarray_support/coordinates.py", line 228, in _index_multiple
    if np.all(values[index] == value):
              ^^^^^^^^^^^^^^^^^^^^^^
ValueError: operands could not be broadcast together with shapes (7,) (8,)
```

Exploration:

If I select a subset where all the pressure values are 35000 or below, I successfully get those levels in the output Zarr. It does not appear to be related to the value difference between, or number of levels between, those requested, but to the absolute value of the levels requested.  See table below for more details.

[...]

PR #506, opened 2026-01-20
fix: coordinate (multiple) indexing - check indices before comparison

aaron-hopkinson:
Replaces subselecting only the index elements smaller than the length with a check they're all smaller to prevent a numpy broadcasting error.

This implementation is similar to the existing [`Coordinate._index_single` case](https://github.com/ecmwf/anemoi-datasets/blob/e7b8d58adb83f9764b2df96280ab41e3facb6147/src/anemoi/datasets/create/sources/xarray_support/coordinates.py#L193).

Related to #498


aaron-hopkinson:
@mo-akm I think this should resolve the issue you were having with selecting subsets of vertical levels. Would it be possible to check that this fix works for you, please?

PR #520, opened 2026-02-02
fix(grib-index): support querying float values

frazane:
When indexing GRIB files using the `anemoi-datasets grib-index`, it might occur that the program fails due to duplicated messages even though they are not really duplicates. This occurs for instance with soil variables, where by default values for `level` are interpreted as integers which leads to values such as 0.01, 0.004, etc. to all be decoded as "0". 

To solve this, one needs to use the `key:type` syntax to decode a key in a specific type, e.g. `level:d`. However, this leads to another error because SQLite does not support column names that contain colons. 

As a simple fix, the proposed solution is to quote the columns.

Extra: since I noticed the "grid definition" support was missing from this source (it's on other GRIB-based sources), I allowed myself to include this change as well, since it's small and should not impact the rest.

---


floriankrb (dismissed):

floriankrb (approved):

MeraX:
Thank you for fixing this!

PR #550, opened 2026-02-24
feat(create): support FDB as base source for accumulate

frazane:
This pull request updates the `fdb` source in the dataset creation pipeline to support interval-based date providers used from the `accumulate` source.

### Summary of changes
* When an `IntervalsDatesProvider` is provided, the method iterates over its intervals, adjusting requests accordingly using the provider's `_adjust_request_to_interval` method.
* The generated requests are now processed through the `factorise_requests` function from the `mars` module, which standardizes and deduplicates the requests before further processing.
---

### DIFF
--- a/docs/building/sources/accumulate.rst
+++ b/docs/building/sources/accumulate.rst
@@ -10,7 +10,7 @@
    - The parameter ``accumulation_period`` has been renamed to ``period``.
    - The source can be now different from ``mars`` (e.g., ``mars``, ``grib-index``)
      it must now be explicitly specified as a nested dictionary under the ``source`` key.
-   - The (optional) available accumulation intervals can now be specified using the ``availability`` key.
+   - The available accumulation intervals must now be specified using the ``availability`` key.
 
 Accumulations and flux variables, such as precipitation, are often
 forecast fields, which are archived for a given base date (or reference


src/anemoi/datasets/create/sources/fdb.py
```diff
@@ -89,15 +92,23 @@ def execute(self, dates: DateList) -> ekd.FieldList:
         ekd.FieldList
             The output data.
         """
-
-        requests = []
-        for date in dates:
-            time_request = _time_request_keys(date, self.offset_from_date)
-            requests.append(self.request | time_request)
+        if isinstance(dates, IntervalsDatesProvider):
```
aaron-hopkinson: I'm not super familiar with the code, but I wonder if `DateList` and `IntervalsDatesProvider` could be polymorphic in some way, so that we use the different objects in the same way (i.e. call the same method) to produce the correct list of requests. (I suspect this is similar to what @floriankrb was thinking when he mentioned the duplication)

aaron-hopkinson: > For this PR I wanted to have the minimal changes to just make it work, but I am happy to discuss how we can improve this via a separate PR.

Of course. I'm in favour of that 🙂 
(I didn't mean my comment as a criticism)

PR #559, opened 2026-03-04
feat: average during building datasets

floriankrb:
##  Additional notes ##


floriankrb:
Do we need this? @jakob-schloer is the only one I know who may use it ?

jakob-schloer:
> Do we need this? @jakob-schloer is the only one I know who may use it?

So far, I have used the `rolling_average` with `open_dataset`. @floriankrb, you've mentioned that this becomes quite inefficient when the window becomes too large. At the moment, I only average 4 steps which works well with the `rolling_average` keyword.

frazane:
Could this not be achieved with #567 ?

maxnessbsc:
Hello,

In DE393 we are interested in averaging during the create/load of the anemoi-dataset from FDB. (Specifically daily averages). 

Is this currently work in progress or can the reduce mentioned in https://github.com/ecmwf/anemoi-datasets/issues/567 be used?

FYI: @jakob-schloer

floriankrb:
Blocked by https://github.com/ecmwf/anemoi-datasets/pull/695 , (we need earthkit-data 1.0 to write the correct metadata). 

Also, this branch need to be updated with the recent changes related to trajectories. `execute()` is not the interface to use.

maxnessbsc:
Hello @floriankrb, thank you for your answer.

I add a bit more of context in case it helps:

In DE393, we are using Destination Earth Climate Digital Twin data to train the climate emulator. Climate DT is retrieved from the data-bridges and downloaded to the MN5 FDB. The data is hourly for the atmosphere and daily for the ocean. DE393 requires daily data. 

When we started building the workflow to generate the datasets, anemoi-datasets didn't support the direct FDB functionality and therefore we developed an alternative solution. Our working solution uses earthkit-data to retrieve the fields from the FDB, and in turn uses earthkit-transform to compute daily averages and healpy for healpix-to-healpix interpolations. Then, intermediate zarr stores are written. The anemoi-part of the pipeline takes these zarr stores and generates the anemoi-dataset.

Lately, I've been looking into possible optimizations and maintenance of the workflow. This has lead me to consider a direct FDB -> anemoi-dataset solution for the needs of DE393. Specifically, the existing parallelization techniques within anemoi-datasets such as the anemoi-datasets load provides good optimizations for dataset generation. However, the computing of daily averages during the dataset building is needed to be able to go in this direction. Additionally, I'm also looking into interpolations during the anemoi-dataset building (though will open a separate issue for this). 

I would be happy to give a hand in this development. Let me know if you have a tentative timeline for this feature and if there is anything I can do to accelerate the process.

Best,

Marvin (from BSC)

FYI: @jakob-schloer

src/anemoi/datasets/create/sources/average.py
```diff
@@ -0,0 +1,99 @@
...
+
+from anemoi.datasets.create.source import Source
+from anemoi.datasets.create.sources import source_registry
+from anemoi.datasets.dates.groups import GroupOfDates
+
+LOG = logging.getLogger(__name__)
+
+
+@source_registry.register("average")
+class AverageSource(Source):
```
aaron-hopkinson, line 29: Would this be better as a filter in anemoi-transform?

floriankrb: It could be done differently yes, perhaps we'll add this as a filter and remove this one later, the migration should be straightforward. In the meanwhile, people need this.

frazane:
Should we not rather consider something like proposed in https://github.com/ecmwf/anemoi-datasets/issues/567?

maxnessbsc:
Thank you very much for prioritizing, let me know if you need some help testing it.

floriankrb:
> Should we not rather consider something like proposed in #567?

I agree, reusing some of the yaml recipi API logic from the `accumulate:` source is a nice idea. See https://anemoi--713.org.readthedocs.build/projects/datasets/en/713/building/sources/average.html, from https://github.com/ecmwf/anemoi-datasets/pull/713

Issue #546, opened 2026-02-20
Create command "--test" flag is ignored


frazane:
The `--test` flag available on commands for dataset creation currently seems to be ignored, as all dates are processed instead of (as it used to be) only the first four.

aaron-hopkinson:
It looks like the functionality was removed in [this PR](https://github.com/ecmwf/anemoi-datasets/pull/326/changes#diff-93e85002305993373ef53368447aff80312c38e16a7aa710a30c135dd8f89051), but was missed in the command line utility (so it doesn't do anything).

After discussing with @b8raoult, I think this could be brought back, but it sounded like there were some limitations with the previous approach (I am not familiar with the code), so it might be worth discussing this with him before we do so.

pontuselmrin:
> Hello! The `--test` flag was removed because it wasn't fit for purpose in it's current form. For example, sometimes people would prefer for test to use a lower resolution as this runs much quicker, but in other cases that would cause the test to fail - for example if there was a mismatch between the resolution of a mask and the resolution of the test dataset.
> 
> Instead I'm going to add now a new command that creates a test yaml from an input yaml, courtesy of [@floriankrb](https://github.com/floriankrb), which allows some customisation. And then `anemoi-datasets create` can be ran on the test yaml. It's a more maintainable solution for us and it's much more explicit what the test dataset contains.

Hi @NatalieZelenka,

Thanks for the context, the test YAML approach sounds reasonable.

I do have an open PR (#570) restoring the --test flag, and it would have been appreciated to hear about this change in direction there rather than finding out via this comment.

PR #491, which removed the functionality, contains no mention of this being intentional. Another maintainer described it as an accidental removal in February, and it was labeled as a good first issue to tackle. There was no way for an external contributor to know the removal was desired, which could have been documented in either in the PR description of #491 or on this issue.

Just sharing this so the next contributor has a smoother experience

NatalieZelenka:
Hello, thanks you for sharing your perspective. Sorry that I didn't reply to #570 first instead of here, I didn't see that PR, but I should have looked at it - my mistake! I agree that's not the experience we would hope for an external contributor and I'll make sure the relevant people see your comment so we can understand how the confusion arose relating to this feature.

PR #570, opened 2026-03-11
fix(create): restore --test flag for dataset creation

pontuselmrin:
## Description

This fix restores the --test flag for `anemoi-datasets create` which was removed during a prior refactor. By using this flag, the dataset that is being created is reduced to 4 dates; the grid is coarsened and any ensembles or constants are trimmed. This allows for quick validation of the create pipeline without large data downloads.

**Bugfix**. The --test flag was removed during a refactor of `create.py`. This restores the flag in the CLI by re-instating the `set_to_test_mode()` function and adds unit tests for its functionality.

#546 

##  Additional notes ##

- Added two unit tests
- The change only affects the init part of the pipeline for creating a dataset; hence multiple GPUs scenario doesn't apply
- The code is reinstated from previous commits, ie I tried to reproduce the previous state.
- The values for grid, constants, ensemble and number of tests in the `set_to_test_mode()` are hard-coded values. Perhaps this could be improved in a future feature

### DIFF
--- a/src/anemoi/datasets/create/config.py
+++ b/src/anemoi/datasets/create/config.py
@@ -16,8 +16,9 @@ from typing import Any
 import yaml
 from anemoi.utils.config import DotDict
 from anemoi.utils.config import load_any_dict_format
-from anemoi.datasets.dates.groups import Groups
 from earthkit.data.core.order import normalize_order_by
+
+from anemoi.datasets.dates.groups import Groups
 
 LOG = logging.getLogger(__name__)
 
--- a/src/anemoi/datasets/create/config.py
+++ b/src/anemoi/datasets/create/config.py
@@ -16,6 +16,7 @@ from typing import Any
 import yaml
 from anemoi.utils.config import DotDict
 from anemoi.utils.config import load_any_dict_format
+from anemoi.datasets.dates.groups import Groups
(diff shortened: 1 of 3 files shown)


src/anemoi/datasets/create/config.py
```diff
@@ -338,20 +340,67 @@ def _prepare_serialisation(o: Any) -> Any:
...
+    num_ensembles = count_ensembles(cfg)
+
+    def set_element_to_test(obj):
+        if isinstance(obj, (list, tuple)):
+            for v in obj:
+                set_element_to_test(v)
+            return
+
+        if isinstance(obj, (dict, DotDict)):
+            if "grid" in obj and num_ensembles > 1:
```
MeraX, line 364: Nice PR, I think it would be useful to restore the simple `--test` option.

Would it be possible to limit the `grid` modification to options that support `grid` values like `"20./20."`? There are Anemoi-Datasets YAML options that use the `grid` key and that should not be modified by the `--test` option. I.e. I would suggest limiting this modification the mars sources only.

aaron-hopkinson:
Thanks for this PR. I'll close this because equivalent functionality has been provided via #639, so I don't think it's needed any longer.

PR #605, opened 2026-04-08
chore: change raise NotImplementedError to raise TypeError

jfdev001:
## Description

Per [python docs: built-in execptions#TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), `TypeError`  should be used to indicate an operation is not supported. This PR replaces the `NotImplementedError`, which would imply one should implement the method, to `TypeError` to more clearly indicate the method is not meant to be implemented at all for children of the read-only class.

https://github.com/ecmwf/anemoi-datasets/issues/602

### DIFF
--- a/src/anemoi/datasets/compat/zarr2.py
+++ b/src/anemoi/datasets/compat/zarr2.py
@@ -22,19 +22,19 @@ class _ReadOnlyStore(zarr.storage.BaseStore):
 
     def __delitem__(self, key: str) -> None:
         """Prevent deletion of items."""
-        raise NotImplementedError()
+        pass
 
     def __setitem__(self, key: str, value: bytes) -> None:
         """Prevent setting of items."""
-        raise NotImplementedError()
+        pass
 
     def __len__(self) -> int:
         """Return the number of items in the store."""
-        raise NotImplementedError()
+        pass
 
     def __iter__(self) -> iter:
         """Return an iterator over the store."""
-        raise NotImplementedError()
+        pass
 
 
 class HTTPStore(_ReadOnlyStore):
--- a/src/anemoi/datasets/compat/zarr2.py
+++ b/src/anemoi/datasets/compat/zarr2.py
@@ -21,12 +21,10 @@ class _ReadOnlyStore(zarr.storage.BaseStore):
     """A base class for read-only stores."""
 
     def __delitem__(self, key: str) -> None:
-        """Prevent deletion of items."""
-        pass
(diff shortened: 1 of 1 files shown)


src/anemoi/datasets/compat/zarr2.py
```diff
@@ -22,11 +22,11 @@ class _ReadOnlyStore(zarr.storage.BaseStore):
 
     def __delitem__(self, key: str) -> None:
         """Prevent deletion of items."""
-        raise NotImplementedError()
+        pass
 
     def __setitem__(self, key: str, value: bytes) -> None:
         """Prevent setting of items."""
-        raise NotImplementedError()
```
floriankrb, line 29: This class is read-only (it is called `class _ReadOnlyStore`), using set_item on a read-only object should fail loudly, so `pass` is not an option.

Arguably, we should raise another exception, most likely one that would not be caught easily.

jfdev001: Thank you for the clarification! The goal is to prevent implementation, so `NotImplementedError` is certainly not a good exception because it implies that one **should** implement the method.

I think perhaps the more appropriate exception, would therefore be `TypeError`. Per [python docs: built-in execptions#TypeError](https://docs.python.org/3/library/exceptions.html#TypeError):

> This exception may be raised by user code to indicate that an attempted operation on an object is not supported, and is not meant to be.

If you're okay with this proposal, I will change the following:

```python
 def __delitem__(self, key: str) -> None:
        raise TypeError("Deleting items is not allowed")

 def __setitem__(self, key: str, value: bytes) -> None:
        raise TypeError("Setting items is not allowed")
```

floriankrb: Good idea, yes

jfdev001:
@floriankrb The CI failure due to 


https://github.com/ecmwf/anemoi-datasets/actions/runs/24395197665/job/71260494246?pr=605#step:6:5416


showed up also in another PR i made (see https://github.com/ecmwf/anemoi-datasets/pull/597#issuecomment-4242720361). I provided what i believe is a fix in that other PR here:



https://github.com/ecmwf/anemoi-datasets/pull/597/changes/5a79d288d143c6de1ec6f5217b63991d69b11ac0

floriankrb:
The failure is not related, so I merged your PR.

And yes, you fix is also relevant, I implemented it there: https://github.com/ecmwf/anemoi-datasets/pull/609

Issue #571, opened 2026-03-12
Adding trajectories to anemoi-datasets


floriankrb:
We are planning to add a trajectories layout to anemoi-datasets, alongside the existing gridded and tabular layouts.

A trajectories dataset stores an explicit step dimension in addition to the existing date/variable/grid dimensions, so that each sample naturally covers a full forecast horizon rather than a single snapshot. This avoids reconstructing step sequences at data-load time, which is error-prone and fragile.

The main features will be:

- a new recipe option (layout: trajectories) to build such datasets
- step-aware reading API (open_dataset(..., steps=...))
- naming convention extension to encode both date and step frequencies

A detailed ADR is being drafted. No breaking changes to existing gridded/tabular datasets are planned.

Feedback welcome — particularly on use cases we may have missed.

floriankrb:
<img width="958" height="399" alt="Image" src="https://github.com/user-attachments/assets/1d08cb19-fcb2-495c-8887-b4e3e196646f" />

floriankrb:
Writing some feedback here because the PR (https://github.com/ecmwf/anemoi-datasets/pull/618) will be closed.

> In my data, the base dates are not regular. Creating multiple datasets could be a workaround. But could you support that, please?
When I split my training data set versus validation data set, the idea of envelope (start= end=), using the validity date is a good idea. I do want to drop the trajectories that are between the training and the validation cutoff time.
This addresses the issue of data overlapping with two valid dates and two different base dates. It doesn't address the issue of the case when you have data from two different runs that have been run from the same base data. In such a case, creating two datasets would be ok.

> Make sure you support ensembles also. (done)

> When using this dataset, I will read the data for each (base date, step). Basically, I would like to flatten the base date dimension and the step dimension to have a unique dimension. This will be done in the data loader in Anemoi training if this is not done in Anemoi dataset.

> I need residuals statistics : statistics of the difference between two datasets. But I understand that is another topic.

> I need tendencies statistics, computed between two steps (with deltas of 1h 3h 6h as usual) (done)

> When joining two trajectories datasets, if the base_dates are shifted and the steps are shifted the other way, make sure we can still join by aligning on the valid dates.

> I don't like the word "base_dates", I would rather use "init_dates" (or "initialisation_dates").

floriankrb:
Additional features, implemented in https://github.com/ecmwf/anemoi-datasets/pull/646:

- [X] tendencies statistics, computed between two steps 
[...]

PR #618, opened 2026-04-24
feat: trajectories datasets

floriankrb:
### Overview

Addresses https://github.com/ecmwf/anemoi-datasets/issues/571

Adds a new `trajectories` layout alongside the existing `gridded` and `tabular` layouts. **No changes to existing recipes or behaviour** — this is purely additive. Status is still quite experimental, API may change; we need to build some of these datasets and have more feedback from users.

Trajectory datasets store a 5D array `(base_dates, variables, ensembles, steps, cells)` with the forecast step dimension encoded explicitly at build time.

### What's new

**Build side:**
- New recipe keys: `base_dates`, `steps`, `output.layout: trajectories`
- `create/arguments.py` — typed date arguments (`ValidDates`, `ForecastDates`, `Intervals`, `ForecastIntervals`)
- `Source` base class dispatches `execute()` to `execute_valid_dates()` / `execute_forecast_dates()` / `execute_intervals()` / `execute_forecast_intervals()` — subclasses override what they support; unsupported paths raise `NotImplementedError` with a clear message
- `sources/accumulate/` — refactored; supports `from-zero` accumulation for trajectories
- `MarsSource` refactored into `mars/` package; implements all 4 dispatch methods
- `ForcingsSource` / `ConstantsSource` — implement `execute_forecast_dates()`, injecting `date`/`time`/`step` metadata onto forcing fields.

**Read side (minimal):**
- `usage/trajectories/` — basic store and subset for opening trajectory datasets
- `usage/dataset.py` — detects trajectory layout, exposes `base_dates`, `steps` properties
- Not yet feature-complete: subsetting, slicing, and concatenation are limited

**Docs:**
- ADR-2: design rationale
- `docs/layouts/trajectories.rst`, additions to `docs/building/` and `docs/using/`

### Still to do

- `GribIndexSource` / `FdbSource` — implement `execute_forecast_dates()` for building trajectories from FDB or GRIB index files
- `RepeatedDatesSource` — trajectory support for constant fields (e.g. LSM repeated across all steps)
- Step 0 support (currently requires step ≥ 1 due to accumulated variable semantics)
- Read-side: full subsetting API, proper date-range filtering, concatenation of trajectory datasets
- `XarraySourceBase` — trajectory support if needed in future
- Date span to compute statistics for trajectory datasets may change slightly in the future 

----
📚 Documentation preview 📚: https://anemoi-datasets--618.org.readthedocs.build/en/618/


floriankrb:
Gathering some feedback here:

- In my data, the base dates are not regular. Creating multiple datasets could be a workaround. But could you support that, please? 
- When I split my training data set versus validation data set, the idea of envelope (start= end=), using the validity date is a good idea. I do want to drop the trajectories that are between the training and the validation cutoff time. 
- This addresses the issue of data overlapping with two valid dates and two different base dates. It doesn't address the issue of the case when you have data from two different runs that have been run from the same base data. In such a case, creating two datasets would be ok.
- Make sure you support ensembles also.

floriankrb:
More feedback

- When using this dataset, I will read the data for each (base date, step). Basically, I would like to flatten the base date dimension and the step dimension to have a unique dimension. This will be done in the data loader in Anemoi training if this is not done in Anemoi dataset.

floriankrb:
@reviewers : The state of this PR is not ready for code review yet, it is here to gather feedback on the global design decisions and concepts involved. A review/comment on a given line is likely to be outdated soon.

floriankrb:
This branch has been updated to resolve conflicts with main, this PR becomes more and more difficult to maintain now.

I am confident this is not breaking any use case in main that we want to support. I think it should be merged as it is.

aaron-hopkinson (approved):

Issue #619, opened 2026-04-27
Feat: caching cutout.


dietervdb-meteo:
Combining two high resolution datasets through a cutout operation takes time to load. E.g. 

```yaml
dataset:
        cutout:
          - dataset: cerra-rr-an-oper-0001-mars-5p5km-1984-2020-3h-v2-rmi.zarr
          - dataset: aifs-ea-an-oper-0001-mars-n320-1979-2024-1h-v2-with-era51.zarr
        min_distance_km: 0
        max_distance_km: 666
        adjust: all
```
takes about 4 minutes to load, as illustrated in graph building
```
2026-04-27 11:34:43 INFO Reading the dataset from {'cutout': [{'dataset': '.../cerra-rr-an-oper-0001-mars-5p5km-1984-2020-3h-v2-rmi.zarr'}, {'dataset': '...aifs-ea-an-oper-0001-mars-n320-1979-2024-1h-v2-with-era51.zarr'}], 'min_distance_km': 0, 'max_distance_km': 666, 'adjust': 'all'}.
2026-04-27 11:39:02 INFO Fitting KNNAreaMaskBuilder with 1142761 reference nodes from "data (cutout)".
```

A user will often do many things based on the same cutout dataset: explore different graphs, various test trainings, inference from dataset. A repeated 4 min load time for all these things is bothersome (in particular for inference!).

Most of the time goes into the calculation of which points of the global dataset get removed in the cutout region, and which are kept by the `max_distance_km` feature. But if a user uses the same cutout many times then this calculation gets repeated over and over again, reproducing the same result.

I would therefore suggest to cache (i.e. write to disk in some tmp folder) the result of the cutout operation. When anemoi-datasets is then asked to perform a cutout it can first check if the result of the needed calculation already exists on disk and only do the costly compution if it is not.

floriankrb:
Yes, this seems like a nice optimisation to have (even if it is not a major feature, and it may be targeting only some specific corner cases).
If somebody wants to implement this, it could be nice to see what the root cause is (if the issue is from the algorithm, the underlying libs, reading from data disk, etc). But @dietervdb-meteo seems to point out that this is compute-bound.

Also, the cache invalidation should be reliable (the filename should have a hash with all that could change (the numpy data used or dataset names and/or the package version numbers, etc.)
Another approach would be to add an explicit parameter `cutout.mask: <file-name>` ?

cathalobrien:
[...]

PR #627, opened 2026-05-06
feat(cutout): user-provided pre-computed cutout masks

frazane:
Related to #619 (not a definitive solution).

Adds a `cache` argument to the `cutout` operation so pre-computed masks can be saved and reused across runs, avoiding the cost of recomputing them every time a cutout dataset is opened (significant for large grids).

Pass a path to a `.npz` file:

```python
ds = open_dataset(
    cutout=[lam_dataset, global_dataset],
    min_distance_km=11,
    adjust="all",
    cache="cutout-masks.npz",
)
```

The first call computes the masks and saves them; later calls load them instead of recomputing. `cache=False` (the default) recomputes without writing anything, and `cache=True` is reserved for a future default cache location and currently raises `NotImplementedError`.

On load, the cutout parameters and a fingerprint of each dataset's coordinate arrays are checked against the cache metadata; a mismatch raises `ValueError` so a stale cache is never used silently. Fingerprinting the coordinates (rather than dataset paths) means a grid change from thinning or masking is detected even when the path is unchanged. The masks are written to a temporary file and moved into place so an interrupted save cannot leave a corrupt file, and the anemoi-datasets version is stored with a warning logged on mismatch.

Changes are in `usage/gridded/grids.py` (the `cache` argument plus save/load helpers), `docs/using/combining.rst` (a new caching section), and `tests/test_data.py` (unit tests).

---

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

----
📚 Documentation preview 📚: https://anemoi-datasets--627.org.readthedocs.build/en/627/

----
📚 Documentation preview 📚: https://anemoi-datasets--627.org.readthedocs.build/en/627/


dietervdb-meteo (dismissed):
Tested and LGTM. Comments maybe a bit verbose, but can be instructive for future refs.

floriankrb:
Great contribution, I think this is a start of a larger feature of using caching for anemoi (which we don't necessarly want to implement now)

_ Some update to the documentation could be nice,
_ In the ".npz", we should have two additional keys: {"version": 1 , and "type": "cutout_mask"}. 
We could also include them in a json string "metadata.json"
_ The json string "_params" could be included in the "metadata.json". 
_ Some params are missing in the "_params"
```
            "axis": self.axis,
            "cropping_distance": self.cropping_distance,
            "neighbours": self.neighbours,
            "min_distance_km": self.min_distance_km,
            "max_distance_km": self.max_distance_km,
```
We also need both dataset names (or dicts).

_ Let's reject cache files that are not ending with ".npz" (this will leave some room to use ".nc" or ".json" or other latter on).
_ Finally what about renaming "mask" to "cache" ? 
`open_dataset(..., mask=)` -> `open_dataset(..., cache=)`

This is what I have in mind:
cache = "file.npz"  (read/save from this file)
cache = False (default) -> Force recompute 
cache = True -> Not Implemented (would read/save to the default cache location, with appropriate hidded filename TBD)

frazane:
Thanks for the feedback @floriankrb, I made a few changes based on your suggestions. Is this what you had in mind?

floriankrb:
Close enough yes, but  I don't see why there is this additional arg `cutout_spec` in the open_dataset, I think it should be removed from __init__. The spec is already fully determined by the arguments that are already there (datasets, axis, cropping_distance, neighbours, min_distance_km, max_distance_km). We could just use them as a key when creating/checking the cache file.

We could also keep things simple and have no check for now, and add it later.

src/anemoi/datasets/usage/gridded/grids.py
```diff
@@ -347,11 +357,96 @@ def __init__(
...
+            self._save_cutout_masks(cache_path)
+
+    def _cutout_masks_params(self) -> dict[str, Any]:
+        """Return the cutout parameters that influence mask generation."""
+        datasets = self.lams + [self.globe]
+        datasets_paths = []
+        for ds in datasets:
+            while not hasattr(ds, "path"):
+                ds = getattr(ds, "forward", ds)
+            datasets_paths.append(ds.path)
```
floriankrb: this part has a few concerns:
 
if a dataset has neither .path nor .forward (e.g. a non-Forwards-derived custom class), getattr(ds, "forward", ds) returns ds itself and the loop never exits.
For examplre, combined subclasses (Cutout, Grids) do have a .forward (it's the first subdataset), so the loop descends into only the first child, silently ignoring the rest.
The other PR about synthetic data will break if we use the path as a key.

Other wrappers like Thinning or spatial Masking change the lat/lon grid while leaving .path and dataset name unchanged. The cache would reuse a mask computed on the full grid for a thinned grid without any mismatch detected. The cache key should fingerprint the actual coordinate arrays.

As claude put it : "_The cache's purpose is to skip recomputing masks that were computed from cutout_mask(lam.latitudes, lam.longitudes, globe.latitudes, globe.longitudes, ...). What the key should fingerprint is the lat/lon arrays themselves, not the storage path_."

src/anemoi/datasets/usage/gridded/grids.py
```diff
@@ -347,11 +357,96 @@ def __init__(
...
+            "type": "cutout_mask",
+            "params": self._cutout_masks_params(),
+        }
+
+    def _save_cutout_masks(self, path: str | Path) -> None:
+        """Save the masks to a .npz file."""
+        masks = {f"lam_{i}": self.masks[i] for i in range(len(self.lams))}
+        masks["global"] = self.global_mask
+        # we can avoid pickling with json.dumps
+        np.savez(path, **masks, metadata=np.array(json.dumps(self._cutout_masks_metadata())))
```
floriankrb: Non-atomic save: if the process is killed during _save_cutout_masks, the .npz is left corrupt. Let's use a tmp file and move it (move is atomic)

src/anemoi/datasets/usage/gridded/grids.py
frazane: Good catch. What do you suggest? Should we hash the coordinate arrays?

floriankrb (approved):

Issue #632, opened 2026-05-18
Overwrite statistic_tendencies only when opening a dataset


anaprietonem:
When calling open_dataset we can currently pass statistics pointing a different anemoi-dataset to overwrite the statistics. However there is no way to control if the overwrite can be applied just to the statistic tendencies or to the whole set. It could be useful to have an extra flag to control that.

````
open_dataset(statistics=A.zarr, statistics_tendencies=B.zarr)

```

Behaviour would be:

- open_dataset(base, statistics=A)                           # both from A (backward compatible)
- open_dataset(base, statistics_tendencies=B)                # only tendencies from B
- open_dataset(base, statistics=A, statistics_tendencies=B)  # mix

PR #633, opened 2026-05-18
feat: datasets tendencies overwrite

anaprietonem:
## Description
Previously `open_dataset(ds, statistics=other)` overrode both `statistics` and `statistics_tendencies` from the same source. This PR adds a `statistics_tendencies=` kwarg so the two can be controlled independently or combined in a single call.

**Changes**
- `Statistics` wrapper now takes two optional overrides (`statistic`, `statistic_tendencies`). Fallback: `.statistics` → `statistic` else forward; `.statistics_tendencies` → `statistic_tendencies` else `statistic` else forward.
- Dispatcher in `dataset.py` pops both kwargs into a single wrapper.
- Mock zarr in tests extended to emit `statistics_tendencies_<delta>_*` arrays; added tests for the tendencies-only and combined forms; existing `test_statistics` extended to check tendencies too.

**API**
```python
open_dataset(base, statistics=A)                            # unchanged: both from A
open_dataset(base, statistics_tendencies=B)                 # new: only tendencies from B
open_dataset(base, statistics=A, statistics_tendencies=B)   # new: mix
```

Fully backward compatible since if not specified the new flag, the code does what it was doing before and the flag has a default of None.

Issue explained here https://github.com/ecmwf/anemoi-datasets/issues/632

https://github.com/ecmwf/anemoi-datasets/issues/632

##  Additional notes ##


src/anemoi/datasets/usage/gridded/statistics.py
```diff
@@ -103,4 +137,7 @@ def get_dataset_names(self, names: set[str]) -> None:
             The set of dataset names.
         """
         super().get_dataset_names(names)
-        self._statistic.get_dataset_names(names)
+        if self._statistic is not None:
+            self._statistic.get_dataset_names(names)
+        if self._statistic_tendencies is not None:
+            self._statistic_tendencies.get_dataset_names(names)
```
aaron-hopkinson, line 147: (Not familiar with this code, but given the name of this method, I'm a bit unsure why it returns `None`... It feels like it should return something...)

src/anemoi/datasets/usage/gridded/statistics.py
```diff
@@ -82,7 +111,12 @@ def forwards_subclass_metadata_specific(self) -> dict[str, Any]:
         Dict[str, Any]
             The metadata specific to the forwards subclass.
         """
-        return dict(statistics=self._statistic.metadata_specific())
+        md: dict[str, Any] = {}
+        if self._statistic is not None:
+            md["statistics"] = self._statistic.metadata_specific()
+        if self._statistic_tendencies is not None:
+            md["statistics_tendencies"] = self._statistic_tendencies.metadata_specific()
```
aaron-hopkinson, line 122: Not entirely sure how this method is used, but `statistic_tendencies` comes from `self._statistics` if `self._statistic_tendencies is None` – should this also have the override? (I suspect no, but just checking)

src/anemoi/datasets/usage/gridded/statistics.py
```diff
@@ -31,31 +31,56 @@ class Statistics(Forwards):
...
+        Dataset providing the override for ``statistics``. Optional.
+    statistic_tendencies : Any
+        Dataset providing the override for ``statistics_tendencies``. Optional.
     """
 
-    def __init__(self, dataset: Dataset, statistic: Any) -> None:
+    def __init__(
+        self,
+        dataset: Dataset,
+        statistic: Any = None,
```
aaron-hopkinson: It feels like `statistic` and `statistic_tendencies` should be of type `Dataset` – although I know that it was `Any` before, so I might be wrong...

aaron-hopkinson (dismissed):
LGTM. A few questions/comments, but nothing major.

(It looks like the test failures must have been caused by something else?)

aaron-hopkinson (dismissed):

src/anemoi/datasets/usage/gridded/statistics.py
anaprietonem: Yes the statistic is the original override that was designed to override both statistics and statistics_tendencies, The new statistic_tendencies just allow the override of the tendencies part

src/anemoi/datasets/usage/gridded/statistics.py
anaprietonem: I haven't looked to much to the logic of the model, but I believe the pattern is aligned to what we have in Forwards.py

src/anemoi/datasets/usage/gridded/statistics.py
anaprietonem: fair, I updated the typing!

aaron-hopkinson (approved):

Issue #631, opened 2026-05-18
Enable specification of multiple files in ODB source


neill-b:
If one specifies a yaml file with an ODB source, then `anemoi-datasets` fails with the following error:
`FileNotFoundError: Input ODB file not found: /some/path/*T0000Z_glu_atovs_odb2`
That is, the wildcard has not been expanded.

This is because the `_execute` routine in the ODB source file:
https://github.com/ecmwf/anemoi-datasets/blob/main/src/anemoi/datasets/create/sources/odb.py#L108
doesn't contain code to expand the file-path. This is done in other routines, for instance in grib:
https://github.com/ecmwf/anemoi-datasets/blob/main/src/anemoi/datasets/create/sources/grib.py#L141

This should be added, and it should be checked that `anemoi-datasets` works correctly with both multiple datasets specified and with the incremental formulation.

aaron-hopkinson:
Closed by #634

PR #575, opened 2026-03-13
feat: pydantic recipes

b8raoult:
## Description

Use Pydandic to load the recipe files, including models that my be defined in sources, filters and other plugins.

To add a pydantic mode to a filter:

```python

class Schema(BaseModel):
   fermi: str = "paradox"
   kardashev_scale: int = 3
   
   
@filter_registry.register("the-great-filter")
class TheGreatFilter(Filter):

      schema = Schema
      
      def __init__(self, fermi, kardashev_scale):
         ....
         
``` 

##  Additional notes ##

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

PR #634, opened 2026-05-18
fix: Allow ODB source to work with multiple input files

neill-b:
## Description
To all the creation of tabular datasets from an set of ODB files. At present, if multiple input files are specified, then the ODB code doesn't expand any wildcards and therefore doesn't find the input files. Additionally, the date-time specification needed by incremental initialisation isn't expanded. This change adds the code to allow these file name expansions to happen, and the resulting list of input files to be cycled over and joined together.

In addition, the tabular creator code requires `self.recipe.build.max_fragment_size`, `self.recipe.build.validate_date_ranges` and `self.recipe.build.max_workers` to be included in the yaml (and fails in a confusing way late in the code if they are not). Therefore, I have given these entries default values in `initialise_dataset` if they are not specified in the yaml.

Finally, the test `test_odb` was failing due to a strange specification in the recipe. This has been fixed (hopefully).

Missing feature from the ODB source code.

#631

##  Additional notes ##
To test the code is now working properly, I have run this code through two cases. Both of these are based on a very simple ingestion of AMSU observations, using a set of ODBs from December 2025. The first is specifying the first day of data, and reading that into a zarr file with a single command: `anemoi-datasets create yamls/atovs_multiple.yaml output/multiple_atovs.zarr`. Inspecting the resultant file gives:
```
$ anemoi-datasets inspect output/multiple_atovs.zarr
2026-05-18 13:25:12 INFO Created date indexing: <anemoi.datasets.date_indexing.bisect.DateBisect object at 0x7f8f2fe91790>
2026-05-18 13:25:12 INFO Initializing ReadAheadWriteBehindBuffer with chunk shape (243148, 69), caching 8 chunks (512.00 MB)
2026-05-18 13:25:12 INFO Initializing ReadAheadWriteBehindBuffer with chunk shape (2796202, 3), caching 8 chunks (512.00 MB)
/data/users/neill.bowler/machine_learning/anemoi-datasets-fork/src/anemoi/datasets/usage/dataset.py:756: UserWarning: Private method _subset is overriden in TabularZarr
  warnings.warn(f"Private method {n} is overriden in {self.__class__.__name__}")
📦 Path          : output/multiple_atovs.zarr
🔢 Format version: 0.15.0

📅 Start      : 2025-12-01 00:00
📅 End        : 2025-12-01 18:00
⏰ Frequency  : None
🌎 Resolution : None
🌎 Field shape: None

📅 Data start : 2025-12-01T00:00:00
📅 Data end   : 2025-12-01T20:59:59
📇 Index size : 75,597

📐 Shape      : 29,154,888 × 69 (7.5 GiB)
💽 Size       : 866.4 MiB (866.4 MiB)
📁 Files      : 138


                                         Statistics                                          
┏━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━┓
┃ Index ┃ Variable                      ┃      Min ┃      Max ┃     Mean ┃    Stdev ┃ Units ┃
┡━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━┩
│     0 │ gen_centre@sat                │      nan │      nan │      nan │      nan │ -     │
│     1 │ general_quality_flag@radiance │      nan │      nan │      nan │      nan │ -     │
│     2 │ initial_level_time@body       │      nan │      nan │      nan │      nan │ -     │
│     3 │ qcflags_info_1dvar@radiance   │        0 │      513 │      103 │      205 │ -     │
│     4 │ qcflags_reject_1dvar@radiance │        0 │ 2.62e+05 │ 1.57e+05 │ 1.28e+05 │ -     │
│     5 │ satellite_identifier@sat      │        5 │        5 │        5 │        0 │ -     │
...
```

Using the same one month of data, I have also used incremental creation to make a file from the whole month of data. Similarly, inspecting this gives a sensible result:
```
$ anemoi-datasets inspect output/combined_atovs.zarr
2026-05-18 13:24:23 INFO Created date indexing: <anemoi.datasets.date_indexing.bisect.DateBisect object at 0x7f0c8798d590>
2026-05-18 13:24:23 INFO Initializing ReadAheadWriteBehindBuffer with chunk shape (243148, 69), caching 8 chunks (512.00 MB)
2026-05-18 13:24:23 INFO Initializing ReadAheadWriteBehindBuffer with chunk shape (2796202, 3), caching 8 chunks (512.00 MB)
/data/users/neill.bowler/machine_learning/anemoi-datasets-fork/src/anemoi/datasets/usage/dataset.py:756: UserWarning: Private method _subset is overriden in TabularZarr
  warnings.warn(f"Private method {n} is overriden in {self.__class__.__name__}")
📦 Path          : output/combined_atovs.zarr
🔢 Format version: 0.15.0

📅 Start      : 2025-12-01 00:00
📅 End        : 2025-12-31 18:00
⏰ Frequency  : None
🌎 Resolution : None
🌎 Field shape: None

📅 Data start : 2025-12-01T00:00:00
📅 Data end   : 2025-12-31T20:59:59
📇 Index size : 2,338,146

📐 Shape      : 894,871,740 × 69 (230 GiB)
💽 Size       : 25.9 GiB (25.9 GiB)
📁 Files      : 3,706


                                        Statistics                                        
┏━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━┓
┃ Index ┃ Variable                      ┃   Min ┃      Max ┃     Mean ┃    Stdev ┃ Units ┃
┡━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━┩
│     0 │ gen_centre@sat                │   nan │      nan │      nan │      nan │ -     │
│     1 │ general_quality_flag@radiance │   nan │      nan │      nan │      nan │ -     │
│     2 │ initial_level_time@body       │   nan │      nan │      nan │      nan │ -     │
│     3 │ qcflags_info_1dvar@radiance   │     0 │      652 │     23.3 │     82.8 │ -     │
│     4 │ qcflags_reject_1dvar@radiance │     0 │ 2.62e+05 │ 2.04e+05 │ 1.09e+05 │ -     │
│     5 │ satellite_identifier@sat      │     3 │        5 │     3.77 │    0.973 │ -     │
```

Therefore, I conclude that the new code is working as expected.


src/anemoi/datasets/create/sources/odb.py
```diff
@@ -485,3 +505,25 @@ def pivot_obs_df(df: pandas.DataFrame, values: list, columns: list) -> pandas.Da
...
+    Any
+        The expanded paths.
+    """
+    for path in paths:
+        cnt = 0
+        for p in glob.glob(path):
+            yield p
+            cnt += 1
+        if cnt == 0:
+            yield path
```
ReubenHill: Given that this is exactly the same as https://github.com/ecmwf/anemoi-datasets/blob/2545ad951e1d3ffbc525efd5e2ff6ab3baf4dc85/src/anemoi/datasets/create/sources/grib.py#L63-L82, could it perhaps be pulled out to a common location?

src/anemoi/datasets/create/sources/odb.py
```diff
@@ -122,19 +125,33 @@ def execute(self, dates: Any) -> pandas.DataFrame:
...
-        return df
+        iso_dates = [d.isoformat() for d in dates]
+
+        # Substitute any dates into the file-path
+        paths = Pattern(self.path).substitute(date=iso_dates, allow_extra=True)
+        # Ensure that the result is always a list
+        path_list = paths if isinstance(paths, list) else [paths]
+        LOG.info(f"Paths: {paths}")
+        df_list = []
+        for fname in _expand(path_list):
```
ReubenHill: Is this meant to be either

```suggestion
        paths = paths if isinstance(paths, list) else [paths]
        LOG.info(f"Paths: {paths}")
        df_list = []
        for fname in _expand(paths):
```

or

```suggestion
        path_list = paths if isinstance(paths, list) else [paths]
        LOG.info(f"Paths: {path_list}")
        df_list = []
        for fname in _expand(path_list):
```
?

Or perhaps the `LOG.info(...)` is not useful since it's not been expanded yet?

ReubenHill: `fname` isn't particularly descriptive. I'd go for `path` or `expanded_path`
```suggestion
        path_list = paths if isinstance(paths, list) else [paths]
        LOG.info(f"Paths: {paths}")
        df_list = []
        for path in _expand(path_list):
```

src/anemoi/datasets/create/sources/odb.py
```diff
@@ -122,19 +125,33 @@ def execute(self, dates: Any) -> pandas.DataFrame:
...
+            LOG.info(f"Working on {fname}")
+            temp_df = odb2df(
```
ReubenHill: I think you can just call `temp_df` `df` here

src/anemoi/datasets/create/tabular/creator.py
```diff
@@ -59,6 +59,20 @@ def initialise_dataset(self, dataset: Dataset) -> None:
         dataset : Dataset
             The dataset object to be initialised with arrays and coordinates.
         """
+        # Check the recipt for mandatory fields in tabular datasets
+        if hasattr(self, "recipe"):
+            if hasattr(self.recipe, "build"):
+                if not hasattr(self.recipe.build, "max_fragment_size"):
+                    LOG.info(
+                        "self.recipe.build.max_fragment_size was not specified, " "defaulting to 268435456 (256MB)"
```
ReubenHill: Why are these info strings split into two strings? I.e. why is this not just

```suggestion
                        "self.recipe.build.max_fragment_size was not specified, defaulting to 268435456 (256MB)"
```
?

ReubenHill: Also this should say `256 MiB` (Mebibyte) if we are being pedantic

src/anemoi/datasets/create/tabular/creator.py
```diff
@@ -59,6 +59,20 @@ def initialise_dataset(self, dataset: Dataset) -> None:
...
+                    )
+                    setattr(self.recipe.build, "max_fragment_size", 268435456)
+                if not hasattr(self.recipe.build, "validate_date_ranges"):
+                    LOG.info("self.recipe.build.validate_date_ranges was not specified, " "defaulting to true")
+                    setattr(self.recipe.build, "validate_date_ranges", True)
+                if not hasattr(self.recipe.build, "max_workers"):
+                    LOG.info("self.recipe.build.max_workers was not specified, " "defaulting to 1")
+                    setattr(self.recipe.build, "max_workers", 1)
```
ReubenHill: Where do these defaults come from? I.e. why are they good choices? Are they specified somewhere?

src/anemoi/datasets/create/sources/odb.py
neill-b: Thanks, I've put that into a separate file.

src/anemoi/datasets/create/sources/odb.py
neill-b: I've got rid of the `Log.info` call, as it was made redundant by the printing of the file name just afterwards.

neill-b: Since we have `self.path` I've avoided this and gone for `expanded_path` instead.

src/anemoi/datasets/create/sources/odb.py
neill-b: Done.

src/anemoi/datasets/create/tabular/creator.py
neill-b: Changed to `MiB`. The strings were split to avoid lines being over 100 characters long, but somehow the new line got lost somewhere along the way. So, I've now split them over two lines.

src/anemoi/datasets/create/tabular/creator.py
neill-b: I found 256MiB as a default somewhere else in the code. Using `True` and `1` were just guesses on my part (which seem to work). Better defaults (or a better way of setting them) may exist.

src/anemoi/datasets/create/tabular/creator.py
neill-b: Ah, I see that the pre-commit hook task has undone the splitting across lines. Hmmmm

src/anemoi/datasets/create/sources/utils.py
```diff
@@ -0,0 +1,33 @@
...
+        List of paths to expand.
+
+    Returns
+    -------
+    Any
+        The expanded paths.
+    """
+    for path in paths:
+        cnt = 0
+        for p in glob.glob(path):
```
dashmore-mo: If we are going to use really big lists and wildcards this could be done with an `glob.iglob` which returns an iterator. Then I think you wouldn't need the counter in the loop either.

src/anemoi/datasets/create/sources/odb.py
```diff
@@ -225,6 +241,9 @@ def odb2df(
             output_odb_path=intermediate_odb_path.name,
             sql_query_string=sql,
         )
+        if os.path.getsize(intermediate_odb_path.name) == 0:
+            LOG.info(f"SQL query returned no results for ODB file at {path} with query: {sql}")
+            return None
```
dashmore-mo, line 239: This `==0` size condition is only true if the `odc sql` file output ODB-2 formatting is binary odb. For "ascii" or "wide" formatted odb a few bytes are written.

dashmore-mo: scratch that - sorry - the `-f odb` that ensures the binary formatting is in the `subselect_odb_using_odc_sql` func (I don't think we put that in there in our hackathon version 😆)

src/anemoi/datasets/create/tabular/creator.py
```diff
@@ -59,6 +59,20 @@ def initialise_dataset(self, dataset: Dataset) -> None:
         dataset : Dataset
             The dataset object to be initialised with arrays and coordinates.
         """
+        # Check the recipt for mandatory fields in tabular datasets
```
dashmore-mo:
```suggestion
        # Check the recipe for mandatory fields in tabular datasets
```

src/anemoi/datasets/create/sources/odb.py
```diff
@@ -122,19 +125,32 @@ def execute(self, dates: Any) -> pandas.DataFrame:
...
-        LOG.info(df)
-        return df
+        iso_dates = [d.isoformat() for d in dates]
+
+        # Substitute any dates into the file-path
+        paths = Pattern(self.path).substitute(date=iso_dates, allow_extra=True)
+        # Ensure that the result is always a list
+        path_list = paths if isinstance(paths, list) else [paths]
+        df_list = []
+        for expanded_path in expand_paths(path_list):
```
aaron-hopkinson: For what it's worth, there's a function `iterate_patterns` (in `create/sources/patterns.py`) that seems to do a lot of the setup here – maybe using that function could reduce some of the duplication. 🙂

src/anemoi/datasets/create/tabular/creator.py
```diff
@@ -59,6 +59,20 @@ def initialise_dataset(self, dataset: Dataset) -> None:
...
+        if hasattr(self, "recipe"):
+            if hasattr(self.recipe, "build"):
+                if not hasattr(self.recipe.build, "max_fragment_size"):
+                    LOG.info(
```
dashmore-mo: Is this log level correct - could log as a `LOG.warning()` as the defaults might lead to unexpected behaviour?

src/anemoi/datasets/create/sources/utils.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import glob
+from typing import Any
+
+
+def expand_paths(paths: list[str]) -> Any:
```
aaron-hopkinson: (Following on from the comment above, we could either use `anemoi.datasets.create.sources.patterns.iterate_pattern`, or rename `_expand` in (the same file) and consider it a public function to reduce the duplication).

src/anemoi/datasets/create/tabular/creator.py
aaron-hopkinson: > I found 256MiB as a default somewhere else in the code.

@b8raoult did some benchmarking on the ECMWF HPFC and found that 128-512MB were optimal on this machine (so settled on 256MB as a reasonable default). For `max_workers` there's [some code to automatically estimate this](https://github.com/ecmwf/anemoi-datasets/blob/2545ad951e1d3ffbc525efd5e2ff6ab3baf4dc85/src/anemoi/datasets/create/tabular/finalise.py#L458), so I think `None` is the best default value. For `validate_date_ranges` I suspect that `True` is a reasonable default.

Re: setting the defaults – there was a recent change that went in to add Pydantic schemas for the config (#575). It looks like these values were missed in that PR (the work might have started in parallel). I think adding the defaults to the schema is probably the "correct" place to add them, rather than here. https://github.com/ecmwf/anemoi-datasets/blob/2545ad951e1d3ffbc525efd5e2ff6ab3baf4dc85/src/anemoi/datasets/create/recipe/build.py#L43

aaron-hopkinson (changes requested):
Thanks for the contribution.

I've added a few small comments Re: reducing code duplication and moving the setting of defaults to the pydantic schema – but the changes seem to address the root problem.

src/anemoi/datasets/create/tabular/creator.py
neill-b: Problem solved by deleting code in favour of pydantic validation.

src/anemoi/datasets/create/tabular/creator.py
neill-b: Done. I've used your defaults and moved the validation to the Build class.

src/anemoi/datasets/create/tabular/creator.py
neill-b: Deleted in favour of using pydantic validation.

src/anemoi/datasets/create/tabular/creator.py
neill-b: Deleted in favour of using pydantic validation.

src/anemoi/datasets/create/sources/odb.py
neill-b: `iterate_patterns` is what I was really looking for. I've switched the ODB reader to use that instead. I've reverted the changes to the grib reader, as it wasn't obvious to me how to unpick that to use `iterate_patterns` instead (although that looks like it would be preferable).

src/anemoi/datasets/create/sources/utils.py
neill-b: See above.

src/anemoi/datasets/create/sources/odb.py
aaron-hopkinson: > I've reverted the changes to the grib reader, as it wasn't obvious to me how to unpick that to use iterate_patterns instead (although that looks like it would be preferable).

No problem. I'd say that's out of scope for this PR anyway. Thanks for making the changes here :)

aaron-hopkinson (approved):
LGTM. Thanks!

PR #635, opened 2026-05-18
feat: add BUFR source

aaron-hopkinson:
## Description
Add a BUFR source – currently supports data from mars only, but should be straightforward to adapt to other sources (e.g. files).

(Code was migrated from AI-DOP project (@a1f1878) with changes limited to those required to get code working with anemoi-datasets).

### DIFF
--- a/src/anemoi/datasets/create/sources/bufr.py
+++ b/src/anemoi/datasets/create/sources/bufr.py
@@ -222,21 +222,17 @@ class BUFRSource(Source):
             raise ValueError(f"Invalid source name: {name}, must be 'mars'")
         return create_source(self.context, {name: config})
 
-    def execute(self, dates: list) -> pd.DataFrame:
-        df_list = []
-        for start_dt, end_dt in dates:
-            ekd_ds = self.source.execute([start_dt, end_dt])
-            if ekd_ds is None:
-                continue
-            bufr_reader = BUFRReader(ekd_ds.path)
-            df = bufr_to_dataframe_parallel(bufr_reader, self.bufr_to_df, self.num_processes)
-            LOG.debug(f"Extracted {len(df)} BUFR rows for date range {start_dt} - {end_dt}")
-            df_list.append(df)
-
-        if not df_list:
-            LOG.debug(f"No BUFR records were extracted for {len(dates)} date range(s)")
+    def execute(self, dates) -> pd.DataFrame:
+        ekd_ds = self.source.execute(dates)
+        if ekd_ds is None:
+            LOG.debug(f"No BUFR data returned for date range {dates.start_range} - {dates.end_range}")
             return pd.DataFrame()
 
-        df = pd.concat(df_list, ignore_index=True)
-        LOG.debug(f"Combined BUFR dataframe contains {len(df)} rows")
+        bufr_reader = BUFRReader(ekd_ds.path)
+        df = bufr_to_dataframe_parallel(bufr_reader, self.bufr_to_df, self.num_processes)
+        LOG.debug(f"Extracted {len(df)} BUFR rows for date range {dates.start_range} - {dates.end_range}")
+
+        if len(df) == 0:
+            return pd.DataFrame()
+
         return df
--- a/src/anemoi/datasets/create/sources/bufr.py
+++ b/src/anemoi/datasets/create/sources/bufr.py
@@ -8,8 +8,14 @@
 # nor does it submit to any jurisdiction.
 
 import logging
+from typing import Annotated
+from typing import Any
+from typing import Literal
 
 import pandas as pd
+from pydantic import BaseModel
+from pydantic import ConfigDict
+from pydantic import Field
 
 from anemoi.datasets.create.source import Source
 from anemoi.datasets.create.sources import create_source
@@ -20,6 +26,32 @@ from .bufr_support.bufr_to_df import BUFRToDataFrame
 from .bufr_support.bufr_to_df import bufr_to_dataframe_parallel
 
 LOG = logging.getLogger(__name__)
+
+
+BUFRCondition = Annotated[list[Any], Field(min_length=2, max_length=2)]
+
+
+class BUFRExtractSchema(BaseModel):
+    """Validation schema for BUFR extraction options."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    per_report: dict[str, str]
+    preselect_msg_header: dict[str, BUFRCondition] | None = None
+    preselect_msg_data: dict[str, BUFRCondition] | None = None
+    datetime_position_prefix: str = ""
+    per_datum: dict[str, dict[str, str | int]] | None = None
+    per_datum_format: Literal["long", "wide"] = "wide"
+
+
+class BUFRSourceSchema(BaseModel):
+    """Validation schema for the `bufr` source in recipes."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    source: dict[Literal["mars"], dict[str, Any]] = Field(..., min_length=1, max_length=1)
+    extract: BUFRExtractSchema
+    num_processes: int = Field(default=1, ge=1)
 
 
 @source_registry.register("bufr")
@@ -160,6 +192,8 @@ class BUFRSource(Source):
                 impactParameter:
                   name: vertco_reference_1
     """
+
+    schema = BUFRSourceSchema
 
     def __init__(self, context, *, source: dict, extract: dict, num_processes: int = 1):
         super().__init__(context)
--- a/src/anemoi/datasets/create/sources/bufr.py
+++ b/src/anemoi/datasets/create/sources/bufr.py
@@ -83,8 +83,9 @@ class BUFRSource(Source):
             BUFR key to use for the canonical ``longitude`` output column.
             Defaults to ``"longitude"``.
         per_report : dict
-            Mapping of BUFR key names to output column names for values that
-            occur once per observation report (subset).
+            Mapping of output column names to BUFR key names for values that
+            occur once per observation report (subset). ``latitude`` and
+            ``longitude`` must not be used as keys here.
         per_datum : dict, optional
             Variables with multiple values per report. Structure depends on
             ``per_datum_format``:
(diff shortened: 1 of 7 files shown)


src/anemoi/datasets/create/sources/bufr.py
```diff
@@ -0,0 +1,185 @@
...
+        super().__init__(context)
+        self.source = self._create_source(source)
+        self.bufr_to_df = BUFRToDataFrame(**extract)
+        self.num_processes = num_processes
+
+    def _create_source(self, source_config: dict):
+        if len(source_config) != 1:
+            raise ValueError("Exactly one source must be specified")
+        name, config = source_config.popitem()
+        if name != "mars":
```
anaprietonem, line 221: Out of curiosity, is this something that will be picked up by pydantic in the future? (ie if someone tries to pass a different source here?)

src/anemoi/datasets/create/sources/bufr.py
```diff
@@ -0,0 +1,185 @@
...
+        df_list = []
+        for start_dt, end_dt in dates:
+            ekd_ds = self.source.execute([start_dt, end_dt])
+            if ekd_ds is None:
+                continue
+            bufr_reader = BUFRReader(ekd_ds.path)
+            df = bufr_to_dataframe_parallel(bufr_reader, self.bufr_to_df, self.num_processes)
+            df.rename(columns={"lat": "latitude", "lon": "longitude"}, inplace=True)
+            df_list.append(df)
+        df = pd.concat(df_list, ignore_index=True)
```
anaprietonem: this could raise an error if df_list is an empty list

src/anemoi/datasets/create/sources/bufr.py
aaron-hopkinson: It should do :)

src/anemoi/datasets/create/sources/bufr.py
aaron-hopkinson: Good spot!
I suspect we should probably return an empty dataframe (and then other parts of datasets can raise if they need to)

src/anemoi/datasets/create/sources/bufr.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def execute(self, dates: list) -> pd.DataFrame:
+        df_list = []
+        for start_dt, end_dt in dates:
+            ekd_ds = self.source.execute([start_dt, end_dt])
+            if ekd_ds is None:
+                continue
+            bufr_reader = BUFRReader(ekd_ds.path)
+            df = bufr_to_dataframe_parallel(bufr_reader, self.bufr_to_df, self.num_processes)
+            df.rename(columns={"lat": "latitude", "lon": "longitude"}, inplace=True)
```
anaprietonem: just curious but why is the renaming required here?

src/anemoi/datasets/create/sources/bufr_support/bufr_reader.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class BUFRMessage:
+    def __init__(self, bid: int):
+        self.bid = bid
+
+    def get_array(self, element: str, typ: type, nsubsets: int, missing_val=np.nan) -> np.ndarray:
+        """Wrapper for codes_get_array to work around the inconsistent handling of arrays in eccodes when data is constant"""
+        try:
+            arr = eccodes.codes_get_array(self.bid, element, typ)
```
anaprietonem, line 25: can't earthki-data read buffr data or is this something very specific to eccdoes?

src/anemoi/datasets/create/sources/bufr_support/bufr_reader.py
```diff
@@ -0,0 +1,105 @@
...
+from contextlib import contextmanager
+
+import eccodes
+import numpy as np
+from gribapi.errors import KeyValueNotFoundError
+
+
+class BUFRMessage:
+    def __init__(self, bid: int):
+        self.bid = bid
```
anaprietonem, line 20: what's bid?

src/anemoi/datasets/create/sources/bufr.py
anaprietonem: yes that's fine, also could it worth just having a logger.debug or logger.info showing the length of the df or that could be too noisy? and rather prefer to catch the error properly downstream?

src/anemoi/datasets/create/sources/bufr.py
anaprietonem: could we have a test? or would that require more work?

anaprietonem (commented):
Very nice, thanks @aaron-hopkinson , left a few questions!

src/anemoi/datasets/create/sources/bufr.py
aaron-hopkinson: No objection to using a logger – people can reduce the log level if it's too noisy. I suspect that debug is probably the right level here

src/anemoi/datasets/create/sources/bufr_support/bufr_reader.py
aaron-hopkinson: It can, but I think @pinnstorm and @pwlean had issues where earthkit-data was a little bit too slow for their needs (I think they discussed this with Sandor). Longer term we should do some benchmarking to quantify the difference (things may have changed since the code was originally written), and then push the "fast" code up to earthkit-data if possible. So this should just be considered "temporary" to add the functionality. :)

src/anemoi/datasets/create/sources/bufr_support/bufr_reader.py
aaron-hopkinson: My guess is "buffer ID", but it's essentially just an eccodes pointer at a message afaik

src/anemoi/datasets/create/sources/bufr.py
aaron-hopkinson: I think "lat" and "lon" come from the bufr files, but having "latitude" and "longitude" are part of the "contract" that tabular sources should fulfil.

src/anemoi/datasets/create/sources/bufr.py
aaron-hopkinson: A test that it dies if someone tries to pass another source config?

aaron-hopkinson: Added something (still need to implement a pydantic schema though)

src/anemoi/datasets/create/sources/bufr.py
aaron-hopkinson: Done (both things)

src/anemoi/datasets/create/sources/bufr_support/bufr_reader.py
anaprietonem: yes that's okey, can we add a comment to say in the future we'd plan to replace the eccodes call with ekd?

src/anemoi/datasets/create/sources/bufr.py
anaprietonem: thanks Aaron, I had a look and resolved the other comments. Once this is done, ping me and I will finish the review!

src/anemoi/datasets/create/sources/bufr_support/bufr_reader.py
aaron-hopkinson: Done (but comment elsewhere - at top of class)

src/anemoi/datasets/create/sources/bufr.py
aaron-hopkinson: Done

aaron-hopkinson:
@pinnstorm Note that I swapped the key, value pairs in `per_report` over (so the key is the column name, the value is what you see in the BUFR) – and I also pulled out "latitude" and "longitude" as specific keyword args to ensure that they're always specified (user doesn't need to remember to add them if they're the same values) – this also removes the need for the rename in some instances.

src/anemoi/datasets/create/sources/bufr.py
aaron-hopkinson: (Removed the rename - now need to pass these values explicitly)

anaprietonem (dismissed):
Thanks a lot @aaron-hopkinson ! great work!

aaron-hopkinson:
@pinnstorm, @anaprietonem – I've confirmed that this PR can successfully create a tabular anemoi-dataset using the config we developed earlier today, as long as it runs with the relevant branch in anemoi-transform: https://github.com/ecmwf/anemoi-transform/pull/296

src/anemoi/datasets/create/sources/bufr_support/bufr_to_df.py
```diff
@@ -0,0 +1,449 @@
...
+                return pd.DataFrame()
+
+            df = self.extract_per_report_df(message, nreports)
+
+            if self.per_datum:
+                df = self.per_datum_processing(message, df, nreports)
+
+            return df
+        except Exception as e:
+            warnings.warn(
```
aaron-hopkinson, line 313: When running the integration test config converted to an anemoi-dataset config, I get some warnings:

```
bufr_to_df.py:313: RuntimeWarning: Unexpected error in message: overflow encountered in cast. Skipping this message.
```

(These might be related to the removal of one of the filters – still need to investigate)

aaron-hopkinson: for context, removal of:
```
- filter:
  longitude: "lambda x: np.isfinite(x)"
  latitude: "lambda x: np.isfinite(x)"
```

aaron-hopkinson: (scratch that – can't be related to the removal of the filter because it happens in the source)

anaprietonem: Is there an easy way to check what message is then one being skipped? (what's not clear for me is if that's a message that's corrupted and can't be transformed or rather it's something to do with the numpy casting to a certain format and having issues with the length of some data?) @pinnstorm have you seen that message when running the DOP dataset generation?

aaron-hopkinson: Just ran this through a debugger, the messages that seem to trip this warning seem to just be full of NaNs. Based on what @pinnstorm said, I don't think this is a big deal :)

anaprietonem: Great then we mark this as resolved, thanks for checking!

anaprietonem (approved):
LGTM!

PR #639, opened 2026-05-27
feat: add `create-test-recipe` workflow for test dataset creation and stabilize missing-date CI tests

### LINKED ISSUE #546: Create command "--test" flag is ignored

NatalieZelenka:
## Description
Closes #546.

Adds a supported test workflow for dataset creation:
1. Run `anemoi-datasets create-test-recipe <recipe.yaml> [<test-recipe.yaml>]`
2. Run `anemoi-datasets create <test-recipe.yaml>` ...

Replacing the previously removed `anemoi-datasets create --test`

#546.

##  Additional notes ##

Main updates:
- Uses a script from @floriankrb as the basis
- Adds the new `create-test-recipe` as a CLI command with `--last-dates` to create a recipe with last dates as well. This will help find quickly when variables are missing for the later part of a recipe.
- Keeps date blocks as start/end/frequency and applies reduction through missing ranges, including support for existing missing entries.
- Reuses shared date utilities from `anemoi.utils.dates` for parsing date/frequency.
- Adds tests 
- Adds docs for the new workflow and missing date examples, including mixed missing date formats (mixture of individual and date ranges).

----
📚 Documentation preview 📚: https://anemoi-datasets--639.org.readthedocs.build/en/639/


floriankrb:
looks good to me.

What happens if the target file already exists? so we want an `--overwrite` option?
Does it work for gridded AND tabular datasets (not required to merge, but if only gridded layout is not supported, there should be an "assert that the layout is gridded" somewhere

We should rebase sometime this branch, and decide if it is worth a ATS needed flag.

aaron-hopkinson (dismissed):

aaron-hopkinson (approved):

PR #640, opened 2026-05-27
feat: renaming a dataset

floriankrb:
Adds a CLI command to rename a dataset on disk

`anemoi-datasets rename old.zarr new.zarr`

**Scope**
- It does not rename it on any catalogue.
- **Local stores only** — remote stores (s3://, etc.) are out of scope for this first version.

**Why:** Datasets are often created with names that don't follow the naming convention. That's fine for training, but the catalogue rejects non-conforming names — which blocks publishing, transferring to other HPCs, and reproducibility. Users need a way to give such a dataset a valid name.

A plain directory rename isn't enough: the dataset name is also embedded in the recipe metadata. And we deliberately assign a new UUID — the renamed dataset is a distinct artefact. Keeping the same UUID would leave two datasets with identical UUIDs but different names, which we don't want, even when the underlying data is identical. Different name → different UUID.

What it does, in order:

- Assigns a new uuid.
- Updates the name embedded in the recipe metadata (recipe and _recipe if present) to the target basename.
- Moves the store to the new path.

Works for both gridded and tabular datasets (only touches store attributes + moves the directory)

-----------

----
📚 Documentation preview 📚: https://anemoi-datasets--640.org.readthedocs.build/en/640/


src/anemoi/datasets/commands/rename.py
```diff
@@ -0,0 +1,128 @@
+# (C) Copyright 2024 Anemoi contributors.
```
HCookie:
```suggestion
# (C) Copyright 2026- Anemoi contributors.
```

src/anemoi/datasets/commands/rename.py
```diff
@@ -0,0 +1,128 @@
...
+            raise ValueError("Both source and target must end in '.zarr'.")
+        if not os.path.exists(source):
+            raise FileNotFoundError(f"Source dataset '{source}' does not exist.")
+        if os.path.exists(target):
+            if not args.overwrite:
+                raise FileExistsError(f"Target '{target}' already exists. Use --overwrite.")
+            LOG.warning(f"Overwriting existing dataset at '{target}'.")
+            shutil.rmtree(target)
+
+        new_name = os.path.splitext(os.path.basename(target))[0]
```
HCookie, line 65: Some of these utility methods may be simplified by using pathlib

src/anemoi/datasets/commands/rename.py
```diff
@@ -0,0 +1,128 @@
...
+    def add_arguments(self, parser: Any) -> None:
+        """Add arguments to the command parser.
+
+        Parameters
+        ----------
+        parser : Any
+            The command parser to which arguments are added.
+        """
+        parser.add_argument("source", help="Path of the dataset to rename (must end in .zarr).")
+        parser.add_argument("target", help="New path for the dataset (must end in .zarr).")
```
HCookie, line 40: By the target being able to be another dir, this could be used as a copy. Could a check if it is the same name make sense to restrict use?

src/anemoi/datasets/commands/rename.py
```diff
@@ -0,0 +1,128 @@
...
+            raise FileNotFoundError(f"Source dataset '{source}' does not exist.")
+        if os.path.exists(target):
+            if not args.overwrite:
+                raise FileExistsError(f"Target '{target}' already exists. Use --overwrite.")
+            LOG.warning(f"Overwriting existing dataset at '{target}'.")
+            shutil.rmtree(target)
+
+        new_name = os.path.splitext(os.path.basename(target))[0]
+
+        # --- 1 & 2: update metadata in place on the source store ---
```
HCookie: Numbers mean what?

src/anemoi/datasets/commands/rename.py
```diff
@@ -0,0 +1,128 @@
...
+            recipe = dataset.get_metadata(key)
+            if recipe is None:
+                continue
+            recipe = self._rename_in_recipe(recipe, new_name)
+            dataset.update_metadata({key: recipe})
+
+        # Track the chain of previous UUIDs (for provenance only).
+        previous_uuids = dataset.get_metadata("previous_uuids", [])
+        if old_uuid is not None:
+            previous_uuids = previous_uuids + [old_uuid]
```
HCookie: How would it ever be None?

```suggestion
        if old_uuid is not None:
            previous_uuids.append(old_uuid)
```

HCookie (dismissed):

src/anemoi/datasets/commands/rename.py
cathalobrien:
```
3. **Moves** the store from the source path to the target path.
```
I think the intention is for rename to be able to function as a copy

src/anemoi/datasets/commands/rename.py
```diff
@@ -0,0 +1,128 @@
...
+
+        dataset.update_metadata(uuid=new_uuid, previous_uuids=previous_uuids)
+
+        LOG.info(f"Renamed '{source}' -> '{target}': uuid {old_uuid} -> {new_uuid}, recipe name -> '{new_name}'")
+
+        # --- 3: move the store ---
+        # Drop the Dataset reference so no file handles/locks are held during the move.
+        del dataset
+        shutil.move(source, target)
```
cathalobrien: we could replace this with os.rename, which would stop people from using rename to copy datasets across different filesystems (there are faster ways to do this e.g. dsync from mpifileutils which is available on most HPCs)

src/anemoi/datasets/commands/rename.py
floriankrb: The intent is to rename, not to copy. Like `mv` in the shell.
If the user wants to copy, they should use another command.

src/anemoi/datasets/commands/rename.py
floriankrb: done.

aaron-hopkinson (approved):

Issue #642, opened 2026-05-31
Add ability to define offset between dates and file names


neill-b:
It's desirable to have the ability to specify an offset for tabular datasets. That is, the observations exist within a given window, and the files are named for the centre of that window. Currently, the code assumes that the file names are derived from the starting dates of each window. This introduces the ability to create an offset, so that the filenames can have dates following start_of_window + offset.

aaron-hopkinson:
Discussed via email, but posting here for visibility in case others have this need.

It's possible to add offsets into the generated path with the `strftimedelta` function, e.g.
`/path/to/data/{date:strftimedelta(-3h;%Y%m%dT%H%MZ)}-parameter.grib`

(similar to how `strftime` is used within paths in configs – but with an extra offset, which goes as the first argument to the function, with the function arguments separated by a semicolon)

PR #643, opened 2026-05-31
feat: Enable an offset to the main date to be specified for start and end dates

neill-b:
## Description
It's desirable to have the ability to specify an offset for tabular datasets. That is, the observations exist within a given window, and the files are named for the centre of that window. Currently, the code assumes that the file names are derived from the starting dates of each window. This introduces the ability to create an offset, so that the filenames can have dates following start_of_window + offset.

New feature

#642 


aaron-hopkinson:
It seems that we already support using time offsets in the generation of file paths:

From https://github.com/ecmwf/anemoi-datasets/issues/642#issuecomment-4611849002:  
> It's possible to add offsets into the generated path with the `strftimedelta` function, e.g.
> `/path/to/data/{date:strftimedelta(-3h;%Y%m%dT%H%MZ)}-parameter.grib`
>
> (similar to how `strftime` is used within paths in configs – but with an extra offset, which goes as the first argument to the function, with the function arguments separated by a semicolon)

Apologies for not pointing you in this direction sooner.

@neill-b, are you happy for me to close this PR in this case?

aaron-hopkinson:
(oops, pushed the wrong button!)

neill-b:
Thanks @aaron-hopkinson. Happy to close, as `{date:strftimedelta(3h;%Y%m%dT%H%MZ)}` is exactly the kind of thing I was looking for.

PR #655, opened 2026-06-08
feat: trajectories missing features 3

floriankrb:
_(AI help: Claude found many potential bugs; I checked and made it fix some of them that were real)_

Fixing many bugs/missing features in 'trajectories' layout, none had an impact on 'gridded' or 'tabular' layouts.
(Not fixing the select issue that will be addressed by https://github.com/ecmwf/anemoi-datasets/pull/656).

- **Ensemble** loading: member number used as a raw index (1-based members were written out of bounds).
- Indexing: open_dataset(..., step=i, ...) with other options would not index step correctly. 
- Recipe validation: **steps must be whole hours,** with a frequency that divides end - start; added safeguards.
- Statistics **corner case**: adde hard error when no trajectory fits the statistics envelope (was silent all-NaN)
- **Misc**: ensemble_dimension metadata aligned with gridded semantics (member count)

----


src/anemoi/datasets/create/recipe/output.py
```diff
@@ -66,9 +66,6 @@ def layout(self) -> str:
     chunking: dict[str, int] = Field(default_factory=lambda: {"dates": 1, "ensembles": 1})
     """The chunking configuration for the output."""
 
-    ensemble_dimension: int = 2
```
mc4117, line 69: why is this now removed? I think these things are used by the metadata in core?

src/anemoi/datasets/create/recipe/statistics.py
```diff
@@ -24,7 +24,7 @@
 class Statistics(BaseModel):
     start: str | int | datetime.datetime | None = None
     end: str | int | datetime.datetime | None = None
-    tendencies: list[str, int] | bool | None = Field(default=True)
```
mc4117, line 27: what is the impact of this default change?

tests/create/trajectories_accumulation.yaml
```diff
@@ -68,38 +68,41 @@ input:
...
-    #            expver: "0001"
-    #            class: od
-    #            grid: 20./20.
-    #            param: [ tp ]
-    #            levtype: sfc
-    #            stream: oper
-    #    - rename:
-    #        param:
-    #          tp: tp_accum_48h
+    # - pipe:
```
mc4117, line 71: should this be commented out or removed?

mc4117 (changes requested):
Thanks for this! I left some comments

src/anemoi/datasets/create/recipe/output.py
floriankrb: no, this is not used anywhere in anemoi-core.

tests/create/trajectories_accumulation.yaml
floriankrb: this illustrates that period can be greater than 24h, but there no need to make the test unnecessarily longer.

mc4117: Could you put a comment on this?

src/anemoi/datasets/create/recipe/statistics.py
floriankrb: good catch. 
This woud make the computation of tendencies 'on' by default. And this is not what was discussed. The expected behaviour is to have them 'off' by default and allow users to request it in there config with `statistics: tendencies: true` (at ECMWF, when creating a datasets, we do set this flag to 'on' on the config)

floriankrb: This is now `default=None)`

mc4117 (approved):
LGTM

PR #662, opened 2026-06-11
feat: Add window and cycle date column option to ODB source

### LINKED ISSUE #661: Add window and "cycle date" column options to ODB reader
There are two issues which have been causing challenges for the use of the ODB reader with the `irregular_to_grid` filter.
1. We would normally label our data using the centre of the DA window. That is an 06 UTC file would contain observations from 03 to 09 UTC. However, since the dates needed to be specified in the date block and in the filter, then it was impossible to get the two to match up without losing some data. A temporary fix was to label the data using the start of the DA window, but this is not ideal.
2. The filter `irregular_to_grid` doesn't see the set of dates which are being used. Therefore, it couldn't work with incremental initialisation, since it can't know which part of the data it has been given. A solution to this is create an additional column where the "cycle date" is stored that `irregular_to_grid` can pick up.

neill-b:
## Description
This change principally updates the ODB reader to include two new features:
1. Add a window option. This allows the reader to grab the data from the input date in question (adding and subtracting the start and end of the window). Without this the dates will be taken from start and end dates of the entire window (i.e. a full 24h if four six-hour cycles are given). When the nominal cycle dates are not at the start of the window of observations then this causes problems.
2. Add an option to create a "cycle date" column. This is useful for the `irregular_to_grid` filter which doesn't know about the dates it is being presented, so that it can find these dates from the data.

I have tested this with a few days of ATOVS data, and I have been able to successfully create gridded ODBs, with and without incremental initialisation.

New features

Closes #661

##  Additional notes ##

In `gridded/creator.py` I have added a change for converting the reporting to a string-cast. This is because the reporting of `extra` dates failed, because some of the elements had already been converted to string, which doesn't have an `isoformat` method. I assume that somehow the report of `missing` dates had already touched some of the underlying data which is being reported, so it was attempting a second conversion of those dates.

There's a bit of changing of the types of the date variables in the ODB reader. In order to add the offsets to the dates, I needed them to be in `datetime` format, which is also required by the `odb2df` method. Therefore, it's better to leave all the variables as `datetime`s, deleting the `iso8601_to_datetime` routine and do the final conversion to string within `odb2df`.

I am minded that using `dates.start_range` and `dates.end_range` doesn't feel like a good way to read in ODB data. Therefore, I am tempted to remove this code path, and create a sensible default for the `window` option (something like `"[-3h,3h)"`). This would clean the code a little and make things a bit more of a logical fit.


aaron-hopkinson:
@neill-b 
> I am minded that using dates.start_range and dates.end_range doesn't feel like a good way to read in ODB data. Therefore, I am tempted to remove this code path, and create a sensible default for the window option (something like "[-3h,3h)"). This would clean the code a little and make things a bit more of a logical fit.

Removing this code path makes sense to me. Maybe instead we should not give it a default and make it a mandatory option, as I suspect this might vary between users?

src/anemoi/datasets/create/sources/odb.py
```diff
@@ -71,6 +74,16 @@ def __init__(
             List of column names - values in these columns will
             be spread across different values of the columns. For instance,
             "observed_value" and "quality_control_value". Defaults to [].
+        window : str, optional
+            The observations window to use around the input dates.
+            For instance "[-3h,3h)" to use observations from 3 hours before
+            to 3 hours after the input dates (excluding the final time).
+            Defaults to None, which means that the window will be determined
+            by the start and end datetimes of the full input dates.
+        cycle_date_column : str, optional
```
aaron-hopkinson: I understand why you've named this as "cycle date", but I wonder if window_date makes clearer that it's a reference data within a window (which may or may not align with e.g. DA cycles)

src/anemoi/datasets/create/gridded/creator.py
```diff
@@ -163,7 +163,7 @@ def check_dates_in_data(dates_in_data, requested_dates):
                 if missing:
                     LOG.error(f"Missing dates: {[_.isoformat() for _ in missing]}")
```
aaron-hopkinson, line 164: Wouldn't missing also be of the same type?

src/anemoi/datasets/create/sources/odb.py
```diff
@@ -120,13 +135,25 @@ def execute(self, dates: Any) -> pandas.DataFrame:
...
+            dt = datetime.datetime.fromisoformat(date)
+            if exclude:
+                return dt + offset + sign * small_time
+            else:
+                return dt + offset
 
         df_list = []
-        for expanded_path, date in iterate_patterns(self.path, dates):
-            LOG.info(f"Working on {expanded_path} {date}")
+        iDate = 0
```
aaron-hopkinson: PEP8 prefers snake_case rather than camelCase for regular variables (so `idate` would be better). That said, you could probably use `enumerate` rather than having to manually increment/decrement this variable.

src/anemoi/datasets/create/sources/odb.py
```diff
@@ -120,13 +135,25 @@ def execute(self, dates: Any) -> pandas.DataFrame:
...
+        start = dates.start_range
+        end = dates.end_range
 
-        start = np.datetime_as_string(dates.start_range)
-        end = np.datetime_as_string(dates.end_range)
+        def _add_dates(date: str, offset: datetime.timedelta, exclude: bool, sign: int) -> datetime.datetime:
+            small_time = np.timedelta64(100, "ms")
+            dt = datetime.datetime.fromisoformat(date)
+            if exclude:
+                return dt + offset + sign * small_time
```
aaron-hopkinson: We could probably derive `sign` from `offset` to simplify the signature a bit

src/anemoi/datasets/create/sources/odb.py
```diff
@@ -214,8 +243,8 @@ def odb2df(
     path = Path(path_str)
 
     # Convert ISO8601 datetimes to YYYYMMDDHHMMSS
```
aaron-hopkinson: (Now they're technically datetime objects, so "ISO8601" isn't required in the comment)

aaron-hopkinson (changes requested):
Thanks for the PR. Functionally the changes look good to me – but there are a few (minor) style issues

src/anemoi/datasets/create/sources/odb.py
neill-b: Yup, that name makes a lot of sense.

src/anemoi/datasets/create/gridded/creator.py
neill-b: Honestly, I don't fully understand why it fails on `extra` and not `missing`. My thought is that it's something like a race condition. If `missing` and `extra` are both pointing to the same underlying date which is altered by the call to `isoformat`, then the second call to `isoformat` in `extra` doesn't have the same `datetime` object to operate on. However, now that I write it down, this sounds implausible. Perhaps the best idea is that I remove this change, replicate the problem and then raise an issue for it specifically?

src/anemoi/datasets/create/sources/odb.py
neill-b: Sure, will do.

aaron-hopkinson (approved):
Thanks for this!

PR #668, opened 2026-06-15
fix: support old datasets

floriankrb:
Users reported breaking changes on "some of the older datasets we commonly use, which are missing resolution (or have resolution=None), and/or missing variable_metadata."

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those, please refer to https://anemoi.readthedocs.io/en/latest/***

### DIFF
--- a/src/anemoi/datasets/usage/forwards.py
+++ b/src/anemoi/datasets/usage/forwards.py
@@ -350,9 +350,9 @@ class Combined(Forwards):
         ValueError
             If the resolutions are not the same.
         """
-        # if d1.resolution is None or d2.resolution is None:
-        #     LOG.warning("One of the datasets has no resolution, cannot check compatibility")
-        #     return
+        if d1.resolution is None or d2.resolution is None:
+            LOG.warning("One of the datasets has no resolution, cannot check compatibility")
+            return
 
         if d1.resolution != d2.resolution:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


havardhhaugen:
Thanks Florian. This solves the example I posted in the slack. However some more complex configs still fail. I split them into separate posts to avoid a massive wall of text.

havardhhaugen:
Cutout where one dataset is missing variables_metadata works
Screenshot from 2026-06-15 13-50-41

havardhhaugen:
Cutout + join where one dataset is missing variables_metadata fails. It looks like this happens because `typed_variables` of the `join`dataset only contains the variables of the one which has metadata?
Screenshot from 2026-06-15 13-53-34
Screenshot from 2026-06-15 13-54-13

havardhhaugen:
I also tested cutout with the global dataset that has metadata and our meps dataset (which by the looks of it has partial metadata), this also fails
Screenshot from 2026-06-15 13-56-20
Screenshot from 2026-06-15 13-57-24

floriankrb:
Superseeded by #669

PR #673, opened 2026-06-18
feat: support empty data for bufr

floriankrb:
_(claude summary of changes)_

  A BUFR source returning no observations for a date yielded a schema-less `pd.DataFrame()`,
  crashing at init. `ValueError: DataFrame is missing columns: {'longitude', 'latitude'}`

  Changes:
  - `bufr_to_df.py`: add `BUFRToDataFrame.empty_frame()` — a row-less frame with
    the declared columns/dtypes (per_report → float32, date → datetime64), derived
    from config alone.
  - `bufr.py`: `execute` returns `empty_frame()` on both empty exits.

  Tested: NOAA-19 MHS recipe with an empty leading range now passes init (22-var
  schema resolved) and load; a mixed range builds normally.

### DIFF
--- a/src/anemoi/datasets/create/sources/bufr.py
+++ b/src/anemoi/datasets/create/sources/bufr.py
@@ -223,16 +223,21 @@ class BUFRSource(Source):
         return create_source(self.context, {name: config})
 
     def execute(self, dates) -> pd.DataFrame:
+        # Return an empty frame that still carries the declared output columns
+        # (known from config alone) so a data-less slot keeps a valid schema and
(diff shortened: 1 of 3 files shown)


aaron-hopkinson (approved):
Change looks good to me. We might want to error more explicitly if the dataframe is empty and the wide format is used, but I think this is fine for now. (It'll only error downstream anyway...)

NB: PR description says
> tabular/creator.py: check_dataset_name passes resolution=None instead of
getattr(minimal_input, ...), which needlessly forced a full pipe run on the
first slot.

I think this one was pulled out into another PR :)

PR #678, opened 2026-06-26
fix: Release rolling average assumption.

jakob-schloer:
## Description
Allow forward-looking rolling-average windows by relaxing the window-start validation in RollingAverage.

Previously the constructor required the window start to be strictly negative (`self.i_start <= 0` → error), forcing every rolling average to include at least one look-back step. This rejected legitimate forward-only windows such as `rolling_average: [0, 4, 'frequency']` (average over the current step plus the next 4), which fail with:
```
ValueError: Window start must be negative, got [0, 4, 'frequency']
```
The check is relaxed from `< = 0` to `< 0`, so a window start of 0 (start = current step) is accepted, while a genuinely invalid positive start (window beginning in the future) is still rejected. 

## Additional notes
- Positive starts (future-only windows) remain rejected.
- The `(0, 0, 'freq')` "no change" case described in the inline comment is now actually reachable.

### DIFF
--- a/src/anemoi/datasets/usage/gridded/rolling_average.py
+++ b/src/anemoi/datasets/usage/gridded/rolling_average.py
@@ -55,7 +55,7 @@ class RollingAverage(Forwards):
         # window = (0, 0, 'freq') means no change
         self.i_start = -window[0]
         self.i_end = window[1] + 1
-        if self.i_start <= 0:
+        if self.i_start < 0:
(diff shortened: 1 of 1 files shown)


floriankrb (approved):

RachelFurner:
It might be worth including a warning when taking 'future' states, to ensure users aren't accidentally 'cheating', by using future info (there are lots of times you might fairly want to do this, but worth a warning so users know).

PR #682, opened 2026-06-29
feat: deprecation warning for group by instead of silent ignore

floriankrb:
the legacy ``dates.group_by`` key has been moved to ``build.group_by``. It was ignored. This PR fixes it and add a deprecation warning.


anaprietonem (approved):
LGTM!

Issue #222, opened 2025-02-28
Feature Request: Recompute cumulated fields when combining datasets with different frequency


mpvginde:
If I understand correctly, the current behavior when combining e.g. a dataset with 6h and 3h temporal frequency is that all the fields at the intersecting dates (i.e. every 6h) are kept without any additional computations.
This means that the adjusted 3h-freq dataset know has a new 6h-frequency, but e.g. the `tp` field still represents 3h-accumulation. 

It would be useful introduce a feature that allows for the automatic recalculation of the cumulated fields, so that the adjusted datasets represent the same accumulations.

I'm willing to give it a try, but I was wondering what information is present about the fields in the dataset.
Does the dataset _know_ which fields are accumulated and is there information about the accumulation period in the dataset?

BR,
Michiel

flyIchtus:
Hello @mpvginde did you get an answer to your questions ? 

I am very interested by the topic too, including contributing to a PR if you want to start one (maybe with the help of Baudouin).

If the cumulation information is anywhere (layman here), it should be in the datasets metadata ?
These can come either from the earthkit Field objects the dataset uses to build the .zarr (when building the .zarr from source data files), or from the .zarr itself after build (should be like `ds = open_dataset(my_zarr) ;  mtd = ds.metadata() ; mtd.keys()`). 

Probably we might want this information to lie in the .zarr anyway, so that we can get it to work on the use-case you describe (merging datasets with different frequencies, when already having them as .zarr's). This might be starting point to either check if cumulation params are in the .zarr, or build a small filter (anemoi-transforms ?) to set them in the recipe.

I'd like to give this a try on local datasets of mine.

Then for the recomputation, do you already have an entrypoint ? I guess [concat.py] (https://github.com/flyIchtus/anemoi-datasets/blob/develop/src/anemoi/datasets/data/concat.py) would do the job, if we can add kwargs to recompute some fields (and then provide the computing routines under [compute] (https://github.com/flyIchtus/anemoi-datasets/tree/develop/src/anemoi/datasets/compute).

mpvginde:
Hi @flyIchtus, 
I don't have an answer yet on whether the metadata has information on accumulated fields (my guess is no).
But I a brief chat with @b8raoult about it. He proposed we start simple.
e.g. If you would want to merge two .zarrs with different frequencies:

> How I would do it:
> - Create a class that adds N consecutive dates
> - Select the accumulations only from the 3h dataset
> - Apply that class
[...]

PR #704, opened 2026-07-15
feat: accumulate variables to new temporal frequency 

SerbanVadineanu:
## Description 

When combining datasets with different temporal frequencies, variables such as tp, ssrd and strd are incompatible due to a different accumulation period. The best solution is to regenerate the dataset with the correct accumulation period, but this can be time-consuming. Here, we provide an alternative option to reaccumulate variables (e.g., 3h to 6h, or 1h to 6h) without recreating the dataset. This is implemented using a new class inspired by the existing RollingAverage functionality. 

Usage 

```
ds = open_dataset( 
   "dataset", 
   frequency="6h", 
   reaccumulate=["tp"],   # any other accumulated variables can be listed too 
) 
```

### Validation 

Figures below show the results of reaccumulating tp from 3h to 6h for the CERRA (Figs. 1–2) and UWC-West (Figs. 3–4) datasets. The distributions of tp_6h and its reaccumulated version match well (Figs 1,3), with differences mainly visibile for very small precipitation amounts (<0.01 mm). Absolute differences between tp_6h and its reaccumulated version are also small, typically between -0.05 to 0.05 mm (CERRA) or -0.1 to 0.1mm (UWC-West). 

<figure>
    cerra_histograms_3way
    <figcaption>Figure 1. Distribution of tp for 3h (blue), 6h (red) and reaccumulated 6h (green) for the CERRA dataset.</figcaption>
</figure>

<figure>
  cerra_histogram
    <figcaption>Figure 2. Histogram of differences between tp_6h and its reaccumulated version for the CERRA dataset. </figcaption>
</figure>

<figure>
  uwcwest_histograms_3way
    <figcaption>Figure 3. Distribution of tp for 3h (blue), 6h (red) and reaccumulated 6h (green) for the UWC-West dataset.  </figcaption>
</figure>

<figure>
  histogram
    <figcaption>Figure 4. Histogram of differences between tp_6h and its reaccumulated version for the UWC-West dataset. </figcaption>
</figure>

This change solves problems related to variable compatibility, such as the one below: 

``` 

  File "/perm/nld4523/venv_jun2026_AG_CUDA13/lib/python3.12/site-packages/anemoi/datasets/usage/forwards.py", line 500, in check_variables_compatibility 

    Variable.check_compatibility( 

  File "/lus/h2resw01/hpcperm/nld4523/anemoi_jun2026/anemoi-transform/src/anemoi/transform/variables/__init__.py", line 367, in check_compatibility 

    raise ValueError(f"Variables are not compatible: {'; '.join(reasons)}") 

ValueError: Variables are not compatible: tp: Periods are not compatible: 6:00:00 vs 1:00:00; ssrd: Periods are not compatible: 6:00:00 vs 1:00:00; strd: Periods are not compatible: 6:00:00 vs 1:00:00 

``` 

This PR is a follow-up of feature request #222 (https://github.com/ecmwf/anemoi-datasets/issues/222).

### DIFF
--- a/src/anemoi/datasets/usage/dataset.py
+++ b/src/anemoi/datasets/usage/dataset.py
@@ -185,6 +185,20 @@ class Dataset(ABC, Sized):
                 Subset(self, self._dates_to_indices(start, end), dict(start=start, end=end))._subset(**kwargs).mutate()
             )
 
+        if "reaccumulate" in kwargs:
+            Reaccumulate = self.usage_factory_load("Reaccumulate")
+
+            if "interpolate_frequency" in kwargs:
+                raise ValueError("Cannot use both `reaccumulate` and `interpolate_frequency`")
+            if "frequency" not in kwargs:
+                raise ValueError("`reaccumulate` requires `frequency` to also be specified")
+
+            variables = kwargs.pop("reaccumulate")
+            frequency = kwargs.pop("frequency")
+            step = self._frequency_to_step(frequency)
+
+            return Reaccumulate(self, step, variables)._subset(**kwargs).mutate()
+
         if "frequency" in kwargs:
             Subset = self.usage_factory_load("Subset")
 
@@ -372,6 +386,30 @@ class Dataset(ABC, Sized):
 
         raise NotImplementedError("Unsupported arguments: " + ", ".join(kwargs))
 
+    def _frequency_to_step(self, frequency: str) -> int:
+        """Convert a frequency string to a step, i.e. the number of dataset
+        timesteps per requested timestep.
+
+        Parameters
+        ----------
+        frequency : str
+            The frequency string.
+
+        Returns
+        -------
+        int
+            The step.
+        """
+        requested_frequency = frequency_to_seconds(frequency)
+        dataset_frequency = frequency_to_seconds(self.frequency)
+
+        if requested_frequency % dataset_frequency != 0:
+            raise ValueError(
+                f"Requested frequency {frequency} is not a multiple of the dataset frequency {self.frequency}. Did you mean to use `interpolate_frequency`?"
+            )
+
+        return requested_frequency // dataset_frequency
+
     def _frequency_to_indices(self, frequency: str) -> list[int]:
         """Convert a frequency string to a list of indices.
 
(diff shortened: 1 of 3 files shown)


floriankrb:
Reaccumulating on the fly could be useful indeed, even if rebuilding is a better solution for performance, as you mention.

I suspect this will give incorect data for datasets such as https://anemoi.ecmwf.int/datasets/aifs-od-an-oper-0001-mars-o96-2016-2023-6h-v2-precipitations?tab=variables where you have accumulation over periods that are not the frequency of the dataset. Is there an assertion to make sure it fails in such case?
image

This could be solved by changing slightly the API to using an explicit accumulation period `reaccumulate = { 'tp' : '6h'}`.
The shortcut `reaccumulate = [ 'tp']` using implicitly the frequency as accumulation period can be useful too.

SerbanVadineanu:
I added an assertion that checks whether the accumulation period of the variable we want to reaccumulate is a proper factor of the frequency we subsample the data set with. This means we now reject, for instance, reaccumulating a variable whose raw values are already 18h accumulations if we ask for a 24h frequency, instead of silently summing multiple overlapping 18h-accumulated timesteps of that variable into a wrong total.

I haven't changed the API to support an explicit per-variable accumulation period, since the current implementation always reaccumulates a variable over the same period as the dataset's resampling frequency. Supporting a different period per variable (e.g. a 12h-accumulated precipitation field inside an otherwise 6-hourly dataset) would need a different block structure than what's here. I don't have a concrete use case for that myself, but happy to work on it if you had one in mind.

SerbanVadineanu:
I realised that, for the use case we envision at KNMI, we need additional functionality in this PR. We want to leverage all the data in a 3-hourly data set as training for a 6-hourly model. Thus, we would need on-the-fly 6-hourly reaccumulation at both t+3h and t+6h. For example, for total precipitation we would need to have tp_6h_reacc[t+3h] = tp_3h[t] + tp_3h[t+3h], and for tp_6h_reacc[t+6h] = tp_3h[t+3h] + tp_3h[t+6h], so, given the overlap in accumulation between the two time steps, we need to compute this with a sliding window. The windowing functionality is inspired again from the RollingAverage class, the difference is that we perform summation instead of averaging.

The code now is modified to support both the initial reaccumulation functionality that is paired with a sub-sampling of the data set and this slinding-window-based approach which we call rolling_accumulate. We introduce the window parameter which, similarly to RollingAverage, is a tuple where the window[0] denotes the number of step before the current index where the window starts, and window[1] denotes the the number of steps after the current index where the window ends. window[2] is a string representing the unit of the step (the only option, 'freq', means the window is in number of time steps in the dataset.)

The API call looks like this now:

- for the initial behaviour nothing has changed
`ds = open_dataset( 
   "dataset", 
   frequency="6h", 
   reaccumulate=["tp"] 
) `
- for the rolling_accumulation
`ds = open_dataset(
    "3h-dataset",
    rolling_accumulate=["tp"],
    window=(-1, 0, "freq")
)`

floriankrb:
This feature is getting more complicated that expected, I would suggest that we focus on accumulating during dataset building, ensuring that the feature you need is available during dataset creation. Then build on this api to have something on the fly if needed. 

The yaml recipe API for the accumulation have been consolidated  in https://github.com/ecmwf/anemoi-datasets/pull/713, I wonder if what is in this branch provides what you need (at build time) or if this is an additional feature?

Issue #700, opened 2026-07-13
Cutout datasets do not work with missing dates


b0ws3r:
**Describe the bug**

The length of the data arrays created from a dataset with missing dates causes issues when creating a cutout dataset later on, due to failing `check_same_lengths`. 
- `ValueError: Cutout: Incompatible lengths: 1421 and 1460` 

We were previously able to train with datasets with missing dates. We upgraded our anemoi-datasets version from 0.5.27 to 0.5.38 recently and regenerated some datasets. Our method of dataset creation is the same. I believe this may have been introduced in #575 . When the dates array is created in new versions it is created with length `all_dates - missing_dates`. This causes the bad length check later when checking compatibility for combining datasets.

We have referenced the page [Managing Missing Dates](https://anemoi.readthedocs.io/projects/datasets/en/latest/using/missing.html#fill-missing-gaps), and did not observe that `fill_missing_dates` was ever called. The shape of the dates array is always all_dates - missing_dates, which means that `check_same_lengths` will never succeed. We  tried to stick that setting in every conceivable place in our yaml recipe to no avail. Does it make sense to make a different issue to make the docs more clear? The internal option fill_missing_dates isn't particularly useful if `save_dataset` isn't also exposed in the python API. 

**Version number**

0.5.38

**To Reproduce**
1. Generate an anemoi-dataset with missing dates using recent anemoi-datasets version (>=0.5.37)
2. Specify a cutout dataset with no missing dates
3. Attempt to train with a cutout dataset. (alternately, create another dataset config for a cutout dataset, and attempt to create the dataset)

**Expected behavior**
Either:
1.  `check_same_lengths` should take into account the missing dates, similar to the resolution to #477 

OR
2. `check_same_lengths` should not be performed for dates, since we already check the dims/shape 

**Stack trace**
```
  File "/usr/local/lib/python3.12/dist-packages/anemoi/training/train/train.py", line 97, in __init__
    self._log_information()
  File "/usr/local/lib/python3.12/dist-packages/anemoi/training/train/train.py", line 411, in _log_information
    num_fc_features = len(self.datamodule.ds_train.data.variables) - len(self.config.data.forcing)
                          ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/functools.py", line 995, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
[...]

PR #716, opened 2026-07-23
fix: fix missing date regression for corner cases such as cutout

floriankrb:
claude-generated fix for #700.  Has been checked carefully  (do not trust ai blindly).

Claude said:

>  Root cause (genuine, not a check-tuning workaround). The regression is on the create side. StartEndDates.values (create/recipe/dates.py) started excluding missing dates from the on-disk time axis, so the stored arrays became
>  all_dates − missing_dates long. That violates a long-standing invariant that missing dates are a subset of the time axis — relied on by three independent consumers:
>  - ZarrWithMissingDates locates missing indices by finding missing_dates inside the dates array → with them excluded, .missing reads back empty and fill_missing_dates silently becomes a no-op (exactly the reporter's "never
>  called / ineffective" observation);
>  - Combined.check_same_lengths (cutout/grids) compares raw _len → a shortened LAM fails against a full-length global;
>  - the legacy DatesProvider.__init__ (dates/__init__.py) explicitly warns on set(missing) - set(values), and the trajectory BaseDates.values deliberately keeps the slots — both confirm the intended contract.
>
>
> The #700 regression was introduced by:
>  0ded992f — "feat: pydantic recipes (#575)" by b8raoult, committed 2026-05-08, first released in v0.5.37 (git describe places it at 0.5.37~7).
>  That commit rewrote the dates provider as a pydantic model and added the missing-date exclusion in the new StartEndDates.values (create/recipe/dates.py), shortening the on-disk time axis to all_dates − missing_dates. The
>  legacy provider it replaced built values as the full range, so before #575 cutout worked. (a32e7d80 / #649 later only optimised date not in self.missing → set membership; it didn't change the behaviour.)
>
>
>  The fix (two parts):
>  1. StartEndDates.values returns the full range again (missing kept as NaN-filled slots); loading still skips them via Groups' Filter(missing), and the load path writes by date-index so missing slots stay NaN. Removed the
>  now-redundant BaseDates.values override.
>  2. Companion fix in _ConstantsCollector.update (create/statistics.py): the restored all-NaN missing rows would otherwise make a genuinely constant forcing (e.g. cos_latitude) look non-constant and poison the partial-stats
>  merge hash. It now drops fully-NaN rows before the constancy check. Only affects datasets that actually contain missing rows.
>
>  Verification: test_missing_range.py (both cases) now asserts the full time axis, the missing_dates metadata, NaN slots, and constant_fields == ["cos_latitude"]. test_run[join] (a non-missing dataset) still matches its
>  reference, and the trajectory/dates suites (40 tests) pass — so non-missing datasets and trajectories are unaffected.
>
>  ⚠️  Action needed from a maintainer — new reference data. This deliberately changes the on-disk layout for datasets with missing dates, so the remote mock-mars-<VERSION> references for the missing-date recipes are now stale
>  and test_run[missing], test_run[missing-range], test_run[missing-mixture] will fail on shape (reference=8 != actual=10) until regenerated. That failure is the fix working, not a bug — the test prints the tar/scp commands to
>  upload fresh references. I left VERSION at 0.19 since bumping it would force regenerating every reference; happy to bump it instead if you'd prefer that signal.


tests/create/test_missing_range.py
```diff
@@ -63,15 +82,26 @@ def test_missing_mixture_dates_are_supported(tmp_path: Path, load_source: LoadSo
...
         np.datetime64("2020-12-30T00:00:00"),
+        np.datetime64("2020-12-30T12:00:00"),  # missing
         np.datetime64("2020-12-31T00:00:00"),
         np.datetime64("2020-12-31T12:00:00"),
         np.datetime64("2021-01-01T00:00:00"),
         np.datetime64("2021-01-01T12:00:00"),
         np.datetime64("2021-01-02T00:00:00"),
         np.datetime64("2021-01-02T12:00:00"),
+        np.datetime64("2021-01-03T00:00:00"),  # missing
         np.datetime64("2021-01-03T12:00:00"),
```
floriankrb, line 99: About the failing tests:  The *missing* reference stores were generated while #575's regression was live (they have all dates minus the missing dates); regenerating them restores the correct full-length time axis that keeps missing dates as NaN slots, matching the correct contract with missing dates included.

This comment can be dismissed when merging.

PR #717, opened 2026-07-23
feat: broadcast-to-trajectories-in-open-dataset

floriankrb:
Allows `open_dataset(traj, gridded, ...)` — any mix, any order — now broadcasts each gridded dataset onto the trajectory layout by valid-time lookup (result[base=b, step=s] = gridded[b+s]) and joins them along the variable axis, preserving argument order. The result behaves like a trajectories dataset.

### DIFF
--- a/src/anemoi/datasets/usage/trajectories/join.py
+++ b/src/anemoi/datasets/usage/trajectories/join.py
@@ -68,7 +68,12 @@ class TrajectoryJoin(Join):
     """
 
     def __init__(self, datasets: list[Dataset], options: Options) -> None:
-        self._template = next(d for d in datasets if not isinstance(d, GriddedAsTrajectory))
+        self._template = next((d for d in datasets if not isinstance(d, GriddedAsTrajectory)), None)
+        if self._template is None:
+            raise ValueError(
+                f"{self.__class__.__name__} requires at least one trajectory member "
+                "(all members are broadcast-gridded)."
+            )
         super().__init__(datasets, options)
 
     # ------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


floriankrb:
To be tested... @mc4117

src/anemoi/datasets/usage/trajectories/join.py
```diff
@@ -0,0 +1,170 @@
...
+    ----------
+    datasets : list of Dataset
+        The datasets to join, in argument order.  Gridded members must already
+        be wrapped as :class:`GriddedAsTrajectory`.
+    options : Options
+        Options for the combined dataset.
+    """
+
+    def __init__(self, datasets: list[Dataset], options: Options) -> None:
+        self._template = next(d for d in datasets if not isinstance(d, GriddedAsTrajectory))
```
HCookie: Probably worth catching and throwing an informative error if no `GriddedAsTrajectory` found?

src/anemoi/datasets/usage/trajectories/join.py
```diff
@@ -0,0 +1,170 @@
...
+    ----------
+    dataset : Dataset
+        The dataset to test.
+
+    Returns
+    -------
+    bool
+        True if the dataset has the 5-D trajectories layout.
+    """
+    return len(dataset.shape) == 5
```
mc4117, line 49: this feels a bit dangerous? do we have a more rigorous check?

floriankrb: Trajectory datasets can be wrapped in Forwards subclasses (subset, select, rename, etc…)  `isinstance(d, TrajectoryStore)` would be wrong.

src/anemoi/datasets/usage/trajectories/join.py
floriankrb: added ` if self._template is None:`

mc4117 (approved):
LGTM! Great work :)

Issue #718, opened 2026-07-24
copy command may fail when grid dimension is large


aamemiya:
**Describe the bug**

The command "anemoi-datasets copy" returned the following error, even though the original dataset had consistent array shapes and chunk sizes for all variables. 
This sometimes happens when the dataset has large number of grid points.  

```
  File "/home/amemiya/miniconda3/envs/anemoi2/lib/python3.12/site-packages/anemoi/utils/cli.py", line 266, in cli_main
    cmd.run(args)
  File "/home/amemiya/anemoi/version/anemoi-datasets-main/src/anemoi/datasets/commands/copy.py", line 701, in run
    copier.run()
  File "/home/amemiya/anemoi/version/anemoi-datasets-main/src/anemoi/datasets/commands/copy.py", line 595, in run
    check_zarr(self.target, self.verbosity)
  File "/home/amemiya/anemoi/version/anemoi-datasets-main/src/anemoi/datasets/misc/check.py", line 113, in check_zarr
    _check_group(zarr.open(path, mode="r"), verbosity, path)
  File "/home/amemiya/anemoi/version/anemoi-datasets-main/src/anemoi/datasets/misc/check.py", line 41, in _check_group
    _check_array(group[name], verbosity, *path, name)
  File "/home/amemiya/anemoi/version/anemoi-datasets-main/src/anemoi/datasets/misc/check.py", line 67, in _check_array
    assert math.prod(array.shape) % math.prod(array.chunks) == 0
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

** Version number **
both 0.5.40 and current main (63dba5f)

**To Reproduce**
- create a dataset which has 38009 grid points 
- run anemoi-datasets copy on that dataset
- the error occurs for latitudes where array.shape =(38009,) and array.chunks=(19005,)

This may also occur for other numbers of grid points as long as it is large enough and odd number.  

**Possible solution**
In copy_array functoin, create_array (create_dataset in older versions) is called without "chunks" parameter for variables other than "data", such as dates and latitudes. When the size of those dimension variables is large, a smaller chunk size may be set automatically, which may not evenly divide the array size. 
The error was resolved by adding the keyword argument "chunks=data.shape".    
https://github.com/ecmwf/anemoi-datasets/blob/main/src/anemoi/datasets/commands/copy.py#L420

floriankrb:
[...]

PR #724, opened 2026-07-29
fix: fix issue 718

floriankrb:
(ai-generated) I reviewed the code, LGTM.

## Root cause:

 copy_array (copy.py:420) copied the auxiliary 1-D arrays (latitudes, longitudes, dates, …) with target.create_array(name, data=data) and no
  chunks. Zarr then auto-guesses a chunk size against a byte budget (~256 KB). For latitudes of shape (38009,) (≈297 KB) it halves the array to
  chunks (19005,), leaving a valid-but-partial final chunk. The zarr-v2 branch of check_zarr (check.py:67) asserts the shape is evenly divisible by
  the chunk, so it aborts. It only bites when the array is large enough to be split (and the length doesn't divide evenly), which is exactly "large,
  odd number of grid points". The main data array is unaffected because copy_data sets its chunks explicitly.

## fix

  Pass chunks=data.shape when creating these arrays, so each small coordinate/dimension array is stored as a single chunk — the chunk count then
  always divides the shape, and the copy no longer depends on zarr's automatic guess:

  target.create_array(name, data=data, chunks=data.shape)

### DIFF
--- a/src/anemoi/datasets/commands/copy.py
+++ b/src/anemoi/datasets/commands/copy.py
@@ -417,7 +417,11 @@ class ZarrCopier:
         data = source[name][...]
         if name in target:
             del target[name]
-        target.create_array(name, data=data)
+        # These are small coordinate/dimension arrays (e.g. latitudes, longitudes,
+        # dates). Store each as a single chunk so the chunking does not depend on
+        # zarr's automatic guess, which can produce a partial final chunk for large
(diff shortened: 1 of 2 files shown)


floriankrb:
This is expected to fix issue #718

aaron-hopkinson (approved):
LGTM

PR #651, opened 2026-06-05
feat: recompute statistics

floriankrb:
Introduce this command : `anemoi-dataset compute <dataset-name> --statistics` , requested by various users.

https://anemoi--651.org.readthedocs.build/projects/datasets/en/651/cli/compute.html#compute-command

Mostly coded by AI, I tested and it looks fine.

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

----
📚 Documentation preview 📚: https://anemoi-datasets--651.org.readthedocs.build/en/651/

PR #730, opened 2026-08-07
feat: output zarr when recomputing statistics

VeraChristina:
## Description
Write outputs to a zarr store when recomputing statistics with `anemoi-datasets compute`

We need to load residual statistics to train downscaling diffusion models. PR https://github.com/ecmwf/anemoi-datasets/pull/651 adds a command to compute residual statistics. PR https://github.com/ecmwf/anemoi-datasets/pull/633 allows to selectively overwrite tendencies statistics from a zarr store with residual statistics using `open_dataset(my_dataset, statistics_tendencies=residual_statistics.zarr)`. This requires a zarr store as input instead of a JSON which the compute command currently produces.

Edit: We also need the dates arrays to be able to access metadata of a dataset opened as described above. I am not sure if adding the dates array is the best approach. Alternatives would be to change how we overwrite statistics or how we access metadata; both of these seem to require more changes.

It is not clear to me whether other use cases need the JSON output. To overwrite statistics, tendencies statistics etc., zarr stores are required. If the intent is to compare statistics against recomputed statistics with `--compare`, then JSON might be the better choice. So we might want to enable both options.

Relates to point 5 of https://github.com/ecmwf/anemoi-core/issues/1274

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-datasets--730.org.readthedocs.build/en/730/

### DIFF
--- a/docs/cli/compute.rst
+++ b/docs/cli/compute.rst
@@ -174,18 +174,6 @@ low-resolution grid and the native low-resolution dataset:
 
     anemoi-datasets compute hi-res grid=o96 --statistics-residual lo-res
 
-Use the residual statistics as an override when opening a dataset for training:
-
-.. code-block:: bash
-
-    anemoi-datasets compute hi-res grid=o96 --statistics-residual lo-res --statistics-tendencies 6h
-    # → produces hi-res.statistics.zarr
-
-.. code:: python
-
-    ds = open_dataset("hi-res",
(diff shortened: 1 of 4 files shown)


VeraChristina:
Closing this as we decided to keep the json output format, and instead enable json input to overwrite the statistics tendencies or statistics or residual tendencies

PR #732, opened 2026-08-10
perf: chunk along grid dimension by default

cathalobrien:
## Description
This PR changes the default chunking used in anemoi-datasets.

Currently each date is 1 file by default. If a single date is too large for a single file, it is split into multiple files along the variable dimension.

I have also added a minimum size of 32MB per date. If a single date is smaller than this, it will be kept as a single file. This is done to save inodes on low res datasets, and because synthetic benchmark results (see below) show a performance degradation from chunking low-res datasets. For context, an o96 global date with 100 variables is 10MB and an n320 global date with 100 variables is 100MB.

This PR changes the default behaviour. Now, each date will be split into 4 files along the grid dimension. If 4 files is not sufficient for a single date, the number of files will be doubled until it fits.

In practice this means n320 and o1280 global grids with 101 variables will have 4 files per date. o96 grids and smaller will have 1 file per date (minimum chunk size). o2560 grids will have 8 files per date

I have added tests for o96, n320, o1280 and o2560 chunking:
```
# dates=1, variables=101, ensembles=1, grid_points=$grid_points
    ("name", "grid_points", "expected_chunks"),
    [
        ("o96", 40_320, 1),
        ("n320", 542_080, 4),
        ("o1280", 6_599_680, 4),
        ("o2560", 26_306_560, 8),
    ],

```

----
📚 Documentation preview 📚: https://anemoi-datasets--732.org.readthedocs.build/en/732/


cathalobrien:
relevant background: some benchmarking I did looking at dataloader throughput with different gpus-per-model counts (determining how many grid points each process reads) and differently chunked datasets (how many grid points per file)
results_dataset_throughput 1

mchantry:
Great idea. I think @aaron-hopkinson or @floriankrb would be better to do a formal review.

src/anemoi/datasets/create/recipe/output.py
```diff
@@ -122,6 +136,32 @@ def get_chunking(self, coords: dict) -> tuple:
...
+            while True:
+                if chunk_bytes <= self._MAX_CHUNK_BYTES:
+                    break
+                if chunks[grid_axis] == 1:
+                    raise ValueError(
+                        f"A single-grid-point chunk requires {chunk_bytes:,} bytes, "
+                        f"exceeding the {self._MAX_CHUNK_BYTES:,}-byte codec limit."
+                    )
+                splits *= 2
+                chunk_bytes = set_grid_chunk_size()
```
MeraX, line 163: I like the idea of this PR! 👍 

I briefly had a look at this algorithm and it appears to me that the chunking derived from the array size. However, keeping in mind that zar compresses the data, the chunk file size will be smaller than the size of the data. So given a target size of 32 MiB and a data of an exact multitude of this size, all chunks would be of sub-ideal size. Would you agree? But overall, this issue might be rather superficial and not significant in reality.

src/anemoi/datasets/create/recipe/output.py
```diff
@@ -41,6 +44,11 @@ class OutputBase(BaseModel):
 class GriddedOutput(OutputBase):
     """Output configuration for gridded datasets."""
 
+    _DEFAULT_GRID_SPLITS: ClassVar[int] = 4
+    # Blosc and several other codecs use signed 32-bit buffer sizes.
+    _MAX_CHUNK_BYTES: ClassVar[int] = 2**31 - 1
+    _MIN_CHUNK_BYTES: ClassVar[int] = 2**25  # 32 MiB
```
MeraX, line 50: could it be useful to make these thresholds configurable in the configuration? Could this help to adjust easily for different file systems with different optimal chunk sizes?

src/anemoi/datasets/create/recipe/output.py
cathalobrien: Hey @MeraX, that's a great point. In my experience, the compression from Zarr is 2x so it is a relevant point. I will do some testing with dummy datasets

src/anemoi/datasets/create/recipe/output.py
cathalobrien: yeah good idea.
The MAX_CHUNK_BYTES is a limitation from the codec library (which is removed in zarr3). I suppose if you wanted to chunk more aggressively, you could decrease that. I'll rework this part and can expose them via a config

MicheleCattaneo:
Nice PR, thanks!
I'd like to share some results on our regional stretched grid model. ERA5-n320 as boundary and a ~1km dataset. The model was a multistep model with 7 inputs and 6 outputs.
We created a tiny version of the regional dataset with 4 chunks along the grid dimension and experiments with sharing over 2 or 4 GPUs, with and without chunked dataset. All the other run parameters were untouched.

Here are some manually-retrieved stats:

- Run 1: https://mlflow.ecmwf.int/#/experiments/456/runs/ce0903946bee4530a750c2744807822c

Dataset: chunked
GPU-per-model: 2
Throughput: 0.11it/s = 9.09 s/it
System mem: ~50%


- Run 2: https://mlflow.ecmwf.int/#/experiments/456/runs/68633aca0d0b49adb47e215007f523bf

Dataset: chunked
GPU-per-model: 4
Throughput: 0.13 it/s = 7.69 s/it
System mem: ~50%



- Run 3: https://mlflow.ecmwf.int/#/experiments/456/runs/a2a9a3e39b2e4defb4c2c571580afc37

Dataset: old
GPU-per-model: 4
Throughput: 0.15 it/s = 6.6 s/it
System mem: ~50%


- Run 4: https://mlflow.ecmwf.int/#/experiments/456/runs/c00455a6287e4c84a544dfb185cb7803

Dataset: old
GPU-per-model: 2
Throughput: 0.12 it/s = 8.35 s/it
System mem: ~50%

Note that these were run on Anemoi versions that are a few month old compared to main. These results are not really what I would have expected. Is is possible that the benefit of a sharded model and a chunked dataset is not present in the SGM setting?

cathalobrien:
@MicheleCattaneo Thanks for testing the setup so quickly! and double-thanks for running with the profiler. I will look at the traces and try say for sure.
It depends how the grid points within the stretched grid are decomposed across GPUs when using model sharding, if most of the grid points end up on one GPU and you have load balancing issues then the benefit of dataset chunking will be lessened because that GPU will need all/most of the chunks.

also might be good to test with both old and new datasets on the same filesystem to control for potential performance impacts there

MicheleCattaneo:
Hey @cathalobrien 
Sorry this was not clear in my message, @frazane created the chunked regional dataset but he did not use your implementation proposed here, he did it a few weeks ago

frazane:
Looked into this and found that `Cutout` was essentially making chunking useless. Here's the fix: https://github.com/ecmwf/anemoi-datasets/pull/736.

@MicheleCattaneo I measured substantial performance improvements, both in terms of memory reduction and training throughput

dietervdb-meteo:
Thanks @MicheleCattaneo and @frazane for checking the regional modelling implications of this, and fixing them! I think now I'm fine with this becoming a default as it seems a pure win.

src/anemoi/datasets/create/recipe/output.py
cathalobrien: updated the min chunk size to reflect compression

src/anemoi/datasets/create/recipe/output.py
cathalobrien: I expose the following values in the recipe now
```
output:
  default_grid_splits
  min_chunk_sizes
  max_chunk_sizes
```

src/anemoi/datasets/create/recipe/output.py
```diff
@@ -122,6 +140,32 @@ def get_chunking(self, coords: dict) -> tuple:
...
+                chunks[grid_axis] = max(1, (grid_size + splits - 1) // splits)
+                return prod(chunks) * np.dtype(self.dtype).itemsize
+
+            chunk_bytes = set_grid_chunk_size()
+            while splits > 1 and chunk_bytes < self.min_chunk_bytes:
+                splits //= 2
+                chunk_bytes = set_grid_chunk_size()
+
+            while True:
+                if chunk_bytes <= self.max_chunk_bytes:
```
japols: can we simplify this to `while chunk_bytes <= ...` ?

japols (changes requested):

src/anemoi/datasets/create/recipe/output.py
cathalobrien: good catch, done

japols (approved):
LGTM

PR #713, opened 2026-07-18
feat: consolidate accumulations and trajectories vocabularies (+support for hindcast)

floriankrb:
This addresses issues that were raised when building accumulated and/or trajectories datasets  (https://github.com/ecmwf/anemoi-datasets/issues/571#issuecomment-5091372240) . Updating the doc and the code to consolidate the vocabularies used in the recipes.

It has no impact on anemoi-datasets usage (i.e. no changes expected for `open_dataset(..)`)

Additionally it implements support for hindcasts data.

The user facing changes are summarized in here: https://anemoi--713.org.readthedocs.build/projects/datasets/en/713/building/sources/accumulate.html 
`anemoi-datasets recipe --migrate` is expected to migrate most of the old recipes.

Arguably this could be split in different PRs.

It is build on top of the branch `/feat/earthkit-data-1.0` and relies on anemoi-transform's branch `feat/earthkit-data-1.0`

- it contains also support for hindcast data (hence the branch name)
- and some memory optimisations for building trajectories.

----
📚 Documentation preview 📚: https://anemoi-datasets--713.org.readthedocs.build/en/713/

PR #734, opened 2026-08-12
feat: support sub-hourly steps for gridded datasets

### LINKED ISSUE #731: Add support for sub-hourly dataset creation
## Summary of the problem

Currently, gridded datasets can't be built at sub-hourly frequency. With `dates.frequency: 10m`
the date list is correct, but archive-facing sources (at least the FBD one) discard anything sub-hourly when building the request (there are hardcoded floorings and casts to integer), so only one field in six is retrieved. In general, in other parts of the code, hourly time steps are assumed (see the "Why it is not possible" below)

## Why we need it

Our use case at MeteoSwiss is the generation of a 10 minutes dataset from the ICON model (reanalysis) for surface variables, clouds and precipitation (no pressure levels). We aim at emulating the ICON model at this high temporal frequency for a "nowcasting" application (without observations, initially).
Allowing sub-hourly datasets does not seem to require a lot of code changes and I believe that other users might be interested in a similar use case.

## Why it's not possible now

A step is expected an integer in hours, when a request is built:

-  In `sources/fdb.py`  we have `_time_request_keys` which divides by 3600, so 00:10–00:50 all end as
  `step=0`
- In `arguments.py`,  both `adjust_request` methods do the same
- In `sources/accumulate/interval_generators.py`, the  steps are parsed with `int()`, turned
  into horuly offsets with `timedelta(hours=step)`
- `sources/accumulate/writers.py` will `raise ValueError(f"Accumulation period must be integer hours, got {hours}")`

## Proposed idea

Work on exactly the points above, where instead of assuming integer values, we use timedeltas.
I am working locally on trying to get this dataset to be built correctly and will soon propose a PR. In the meantime, I am happy to discuss!

MicheleCattaneo:
## Description

Currently, it is not possible to generate sub-hourly gridded datasets (at least those with an FDB source).
This PR proposes the changes that allowed me to generate a 10 minute dataset. The result still needs to be verified thoroughly, but the creation finishes without errors or missing values, and the result looks "visually" good.

Existing code treats a step as an integer number of hours in several places, so anything finer is lost. 
This PR adds a pair of helpers that convert between `timedelta` and the archive's step syntax (`10m`, `10h10m`)

This PR is a new feature without breaking changes, since a simple integer still means hours, existing recipes should be unaffected.

Closes #731 

##  Additional notes ##


aaron-hopkinson (approved):
I'm not an expert on this code, but the code changes look reasonable to me. I'm not sure if there's anywhere else that would need changing or what the impacts of the change would be – but the changes look good to me.

src/anemoi/datasets/create/intervals.py
```diff
@@ -7,7 +7,65 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+import re
 from datetime import datetime
+from datetime import timedelta
+
+
+def parse_mars_step(step: str | int | timedelta) -> timedelta:
```
mchantry: Why is it called "mars_step" ?

MicheleCattaneo: Perhaps it's not accurate

The reason was that one of the places where this function was called is here https://github.com/ecmwf/anemoi-datasets/blob/9ed3a466e4bd33900a9d95722ecca0f0950982e3/src/anemoi/datasets/create/arguments.py#L324 where the doc strings mention MARS https://github.com/ecmwf/anemoi-datasets/blob/9ed3a466e4bd33900a9d95722ecca0f0950982e3/src/anemoi/datasets/create/arguments.py#L335 

Should I remove the specific mention to MARS?

mchantry: Thanks for explaining (and sorry for the slow reply).
I leave it to @floriankrb or @b8raoult to weigh in on the naming. It might be that we should pass through the code stripping out some of the historical references to MARS/FDB syntax that are less apt now that anemoi datasets supports a wider range of sources.

src/anemoi/datasets/create/arguments.py
```diff
@@ -335,10 +337,10 @@ def adjust_request(self, interval: Any, request: dict) -> tuple[Any, dict, int]:
         Returns
         -------
         tuple
-            ``(valid_time, adjusted_request, step_hours)``
+            ``(valid_time, adjusted_request, step)``
         """
         r = request.copy()
-        step = int((interval.max - interval.base).total_seconds() / 3600)
+        step = format_mars_step(interval.max - interval.base)
```
aaron-hopkinson: Ah, I've just looked into the failing tests a bit more:

Due to the way the tests are implemented, changing the type here changes the hash that's calculated in `test/create/utils/mock_sources.py`, which causes a test failure (see GitHub action output for details, there are some 404 errors).

There are a few potential fixes:
1. we could return ints from `format_mars_step` when its a whole hour, and only strings for subhourly
2. we could do some normalisation in the mock's filename calculation
3. we could reupload the KGO data with the new hashes

aaron-hopkinson (changes requested):
Some of the tests fail - please see comment for options for potential fixes

src/anemoi/datasets/create/arguments.py
aaron-hopkinson: See #739 for PR which implements option 1 and should fix the tests...

MicheleCattaneo: I think option one sounds good, thanks for looking into it. I will review the PR asap!

floriankrb:
Regarding supporting subhourly datasets (this PR): a lot of the work is already done on earthkit-1.0 and `feature/hindcasts`, by accident. The earthkit-1.0 rework replaced the integer-hour plumbing with timedelta and earthkit components, so the field writers, the traj_point cube key, the on-disk step axis, statistics, the naming regex, and the recipe frequency type are sub-hourly-clean already.
What remains is some leftover `int(.../3600)` plus some deliberate whole-hour rejections. 

Reimplementing this feature `feat/sub-hourly-steps-gridded` on top of the branch `feature/hindcasts` may be an easier path, since this feature will need to be updated anyways (see related comment in https://github.com/ecmwf/anemoi-datasets/pull/713#issuecomment-5490978351)

MicheleCattaneo:
Hi, thanks for this point.

A few questions:

- Do you suggest rebasing `feat/sub-hourly-steps-gridded` onto `feature/hindcasts` and still target main, or merging this one into `feature/hindcasts`?
- Is there a timeline for `feature/hindcasts` to reach main?
- A point on dependencies: anemoi-transform on main requires earthkit-data<1 (and earthkit-meteo<1), so we're on earthkit 0.x because of that. We're about to build a large 10-minute dataset and would prefer a commit on main to produce it. If we target the `feat/earthkit-data-1.0` branch (`feature/hindcasts` targets it)  with the changes needed to produce 10 min data, we would be blocked by anemoi-transform's dependencies.

This PR is rather short and backwards-compatible. Considering the last point especially, would you consider merging this in main, with the rework to match the other PRs moved to later?

floriankrb:
You are correct about anemoi-transform; this one also has its earthkit-1.0 branch that should also be merged.

> Is there a timeline for feature/hindcasts to reach main?

I am targeting the end of september to merge earthkit1.0, but other people are involved and have various priorities.

>  would you consider merging this in main...

Sure, I don't want to have you wait for anything if you can have code running and are ready to build a dataset, and this feature is clearly a core feature that should be supported. So yes, I believe this PR could go into main, even if it is reimplemented.

floriankrb (approved):

MicheleCattaneo:
Should we mention somewhere that 10minute frequencies are supported with this PR (maybe somewhere in the docs), or is it fine like this?

floriankrb:
This looks good and can be merged as far I as know.

I reimplemented it in the branch `feature/hindcasts` (commit d163c5f), trying to mimic what is done here.
@MicheleCattaneo if time permits, could you also test to ensure the feature is there and gives consistent behaviour (i.e. can build your 10m dataset) ? (notice the command `anemoi-datasets recipe --migrate` might be useful to update the recipe)

aaron-hopkinson (approved):

PR #736, opened 2026-08-16
perf(cutout): push the grid index down to the source datasets

frazane:
## Description

`Cutout.__getitem__` loaded the full grid from every source dataset and applied the requested grid index in memory, so a grid-subset request (e.g. a reader rank fetching its shard in model-sharded training) still read and decompressed every chunk of every source. This change pushes the grid index down to the sources, the same way `GivenAxis._get_tuple` already does for `grids`: the request is split across the masked sources with `length_to_slices`, each source is asked only for the bounding window of its selected points, and in-memory re-indexing happens only where the mask has holes inside that window. For a single-LAM cutout the LAM request reduces to a pure slice, so only the intersecting chunks are read. Results are byte-identical.

Performance: makes spatial chunking effective under `cutout`, so model-sharded training reads only each rank's grid shard instead of the full field. Full-grid reads and all other access patterns are unchanged.

—

## Additional notes ##

Tested in a parallel setting on a 1 km stretched-grid configuration (2 nodes, 4 GPUs per model, each reader rank requesting its grid shard): per-step training losses are identical to main, a rank's shard read is 5x faster, training throughput is 1.6x higher (data loading no longer stalls the steps), and dataloader-worker memory is reduced by 50%. The new unit test covers a hierarchical two-LAM cutout with holed masks: equality with the mask-then-concatenate reference across index patterns, and sources are never asked for more than the bounding window.

One known limitation, to be addressed separately: reads are chunk-granular, and shard boundaries generally do not coincide with chunk boundaries, so a rank may still decompress more than it needs when the store has few spatial chunks. Mitigations (finer spatial chunking at dataset build time, or chunk-aligned shard sizes on the training side) are outside the scope of this PR.

---

### DIFF
--- a/src/anemoi/datasets/usage/gridded/grids.py
+++ b/src/anemoi/datasets/usage/gridded/grids.py
@@ -596,10 +596,19 @@ class Cutout(GridsBase):
             index = (index, slice(None), slice(None), slice(None))
         return self._get_tuple(index)
 
+    @cached_property
+    def _mask_indices(self) -> list[NDArray[Any]]:
+        """Source-space indices retained by each mask (LAMs first, then the global dataset)."""
+        return [np.flatnonzero(mask) for mask in self.masks] + [np.flatnonzero(self.global_mask)]
+
     @debug_indexing
     @expand_list_indexing
     def _get_tuple(self, index: TupleIndex) -> NDArray[Any]:
         """Helper method that applies masks and retrieves data from each dataset according to the specified index.
+
+        The grid index is pushed down to the underlying datasets, so only the
+        requested region is read (e.g. only the zarr chunks it intersects),
+        instead of loading full fields and slicing in memory.
 
         Parameters
         ----------
@@ -612,16 +621,21 @@ class Cutout(GridsBase):
             Concatenated data array from all datasets based on the index.
         """
         index, changes = index_to_slices(index, self.shape)
-        # Select data from each LAM
-        lam_data = [lam[index[:3]] for lam in self.lams]
+        slices = length_to_slices(index[self.axis], [len(kept) for kept in self._mask_indices])
 
-        # First apply spatial indexing on `self.globe` and then apply the mask
-        globe_data_sliced = self.globe[index[:3]]
-        globe_data = globe_data_sliced[..., self.global_mask]
+        parts = []
+        for dataset, kept, local in zip(self.datasets, self._mask_indices, slices):
+            if local is None:
+                continue
+            sel = kept[local]
+            window = slice(int(sel[0]), int(sel[-1]) + 1)
+            part = dataset[index[:3] + (window,)]
+            if len(sel) != window.stop - window.start:
+                # the mask has holes (or the slice a step) inside the window
+                part = part[..., sel - window.start]
+            parts.append(part)
 
-        # Concatenate LAM data with global data, apply the grid slicing
-        result = np.concatenate(lam_data + [globe_data], axis=self.axis)[..., index[3]]
-
+        result = np.concatenate(parts, axis=self.axis)
         return apply_index_to_slices_changes(result, changes)
 
     def collect_supporting_arrays(self, collected: list[Any], *path: Any) -> None:
--- a/src/anemoi/datasets/usage/gridded/grids.py
+++ b/src/anemoi/datasets/usage/gridded/grids.py
@@ -621,18 +621,27 @@ class Cutout(GridsBase):
             Concatenated data array from all datasets based on the index.
         """
         index, changes = index_to_slices(index, self.shape)
-        slices = length_to_slices(index[self.axis], [len(kept) for kept in self._mask_indices])
+        # resolve the requested grid index for each dataset
+        # relative to each dataset's own masked points
+        requested_per_dataset = length_to_slices(
+            index[self.axis], [len(source_indices) for source_indices in self._mask_indices]
+        )
 
         parts = []
-        for dataset, kept, local in zip(self.datasets, self._mask_indices, slices):
-            if local is None:
+        for dataset, source_indices, requested in zip(self.datasets, self._mask_indices, requested_per_dataset):
+            if requested is None:
+                # this dataset contributes no columns to the requested grid range
                 continue
-            sel = kept[local]
-            window = slice(int(sel[0]), int(sel[-1]) + 1)
+            # translate the requested positions from "retained points within a mask"
+            # back to native grid-point indices
+            selected_source_indices = source_indices[requested]
+            # read only the contiguous native-grid range spanning the selected points,
+            # instead of loading the whole grid and slicing in memory
+            window = slice(int(selected_source_indices[0]), int(selected_source_indices[-1]) + 1)
             part = dataset[index[:3] + (window,)]
-            if len(sel) != window.stop - window.start:
-                # the mask has holes (or the slice a step) inside the window
-                part = part[..., sel - window.start]
+            if len(selected_source_indices) != window.stop - window.start:
+                # the mask has gaps (or the index has a step) inside the window: drop the extra points
+                part = part[..., selected_source_indices - window.start]
             parts.append(part)
 
         result = np.concatenate(parts, axis=self.axis)
--- a/src/anemoi/datasets/usage/gridded/grids.py
+++ b/src/anemoi/datasets/usage/gridded/grids.py
@@ -621,6 +621,15 @@ class Cutout(GridsBase):
             Concatenated data array from all datasets based on the index.
         """
         index, changes = index_to_slices(index, self.shape)
+
+        reverse = index[self.axis].step < 0
+        if reverse:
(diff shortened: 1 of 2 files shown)


cathalobrien:
nice find, good to see that you get a speedup from chunking and you're no longer dataloader bound

src/anemoi/datasets/usage/gridded/grids.py
```diff
@@ -612,16 +621,21 @@ def _get_tuple(self, index: TupleIndex) -> NDArray[Any]:
...
-        # Concatenate LAM data with global data, apply the grid slicing
-        result = np.concatenate(lam_data + [globe_data], axis=self.axis)[..., index[3]]
-
+        slices = length_to_slices(index[self.axis], [len(kept) for kept in self._mask_indices])
+
+        parts = []
+        for dataset, kept, local in zip(self.datasets, self._mask_indices, slices):
+            if local is None:
+                continue
+            sel = kept[local]
```
cathalobrien: Nice PR, I haven't run the code only looked over it. The idea is good but the code is a little unclear at least to me, when reviewing I tried to add some comments and rename variables to make it clear for myself. Maybe you want to add some of these comments (assuming they make sense)
```python
        # resolve the requested grid index for each dataset
        # relative to each dataset's own masked points.
        requested_per_dataset = length_to_slices(index[self.axis], [len(source_indices) for source_indices in self._mask_indices])

        parts = []
        for dataset, source_indices, requested in zip(self.datasets, self._mask_indices, requested_per_dataset):
            if requested is None:
                # this dataset contributes no columns to the requested grid range
                continue
            # translate the requested positions from "retained points within a mask" 
            # back to native grid-point indices
            selected_source_indices = source_indices[requested]
            # read only the contiguous native-grid range spanning the selected points,
            # instead of loading the whole grid and slicing in memory
            window = slice(int(selected_source_indices[0]), int(selected_source_indices[-1]) + 1)
            part = dataset[index[:3] + (window,)]
            if len(selected_source_indices) != window.stop - window.start:
                # the mask has gaps (or the index has a step) inside window: drop the extra points
                part = part[..., selected_source_indices - window.start]
```

tests/test_data.py
```diff
@@ -1762,6 +1762,54 @@ def test_cutout_masks_version_mismatch_warns(tmp_path, caplog):
...
+    obj.datasets = [*lams, globe]
+    obj.masks = [np.tile([True, False], 5), np.tile([True, True, False, True], 2)]
+    obj.global_mask = np.tile([False, True, True], 4)
+
+    # Reference semantics: mask each dataset, concatenate, then index.
+    masks = [*obj.masks, obj.global_mask]
+    reference = np.concatenate([d.data[..., m] for d, m in zip(obj.datasets, masks)], axis=3)
+    assert obj.shape == reference.shape  # kept points: 5 + 6 + 8
+
+    indices = [
```
cathalobrien: this could be a pytest parameterisation instead of a for loop, and obj could be a property or whatever so its not rebuilt for each parameter.

tests/test_data.py
```diff
@@ -1762,6 +1762,54 @@ def test_cutout_masks_version_mismatch_warns(tmp_path, caplog):
...
 
 
+def test_cutout_get_tuple_pushes_grid_index_down():
+    class _ArrayDataset:
+        def __init__(self, data):
+            self.data = data
+            self.shape = data.shape
+            self.grid_requests = []
+
+        def __getitem__(self, index):
```
cathalobrien, line 1776: I would add a comment add that this class keeps a record of accesses. and then your pytest consists of checking which dataset has been accessed when

tests/test_data.py
```diff
@@ -1762,6 +1762,54 @@ def test_cutout_masks_version_mismatch_warns(tmp_path, caplog):
...
+
+    indices = [
+        (slice(None), slice(None), slice(None), slice(None)),
+        (slice(1, 3), slice(0, 2), slice(None), slice(2, 14)),  # straddles all three datasets
+        (slice(None), slice(None), slice(None), slice(0, 17, 3)),
+        ([0, 2], slice(None), slice(None), slice(4, 9)),
+        (slice(2, 3), slice(None), slice(None), slice(18, 19)),
+    ]
+    for index in indices:
+        np.testing.assert_array_equal(obj[index], reference[index])
```
cathalobrien: shouldn't you check the grid accesses here?

cathalobrien (changes requested):

src/anemoi/datasets/usage/gridded/grids.py
```diff
@@ -612,16 +621,30 @@ def _get_tuple(self, index: TupleIndex) -> NDArray[Any]:
...
-        # Concatenate LAM data with global data, apply the grid slicing
-        result = np.concatenate(lam_data + [globe_data], axis=self.axis)[..., index[3]]
+        # resolve the requested grid index for each dataset
+        # relative to each dataset's own masked points
+        requested_per_dataset = length_to_slices(
+            index[self.axis], [len(source_indices) for source_indices in self._mask_indices]
+        )
 
+        parts = []
+        for dataset, source_indices, requested in zip(self.datasets, self._mask_indices, requested_per_dataset):
```
japols, line 640: we should make this loop here a bit more robust, including handling of non-contiguous, or potentially descending index selections

src/anemoi/datasets/usage/gridded/grids.py
```diff
@@ -612,16 +621,30 @@ def _get_tuple(self, index: TupleIndex) -> NDArray[Any]:
...
+            # read only the contiguous native-grid range spanning the selected points,
+            # instead of loading the whole grid and slicing in memory
+            window = slice(int(selected_source_indices[0]), int(selected_source_indices[-1]) + 1)
+            part = dataset[index[:3] + (window,)]
+            if len(selected_source_indices) != window.stop - window.start:
+                # the mask has gaps (or the index has a step) inside the window: drop the extra points
+                part = part[..., selected_source_indices - window.start]
+            parts.append(part)
+
+        result = np.concatenate(parts, axis=self.axis)
```
japols: we should guard against empty selections here where np.concat([]) would fail

japols (changes requested):
very nice! I like that this preserves the indexing behaviour (concat grids, then slice).

In the future we could also think of extending this to indexing with different grid indices per dataset (slice grids, then concat) which should align even better with grid-chunked datasets. Thoughts @floriankrb ?

also left some comments.

cathalobrien:
i did some benchmarking, and confirm that this PR leads to improved dataloader throughput when using datasets chunked along the grid dim

my benchmark setup was the dataloader benchmark tests used in the integration tests, modified to use the sharded reading logic. with this i can set an arbitrary read group size, and simulate reading at different gpus_per_model. the graphs show **dataloader** throughput on the y axis (higher is better) and simulated gpus-per-model on the x axis. I chunked the local dataset at 1,2,48,16 grid chunks per date. the global dataset was fixed at 1 grid chunk per date. 

here is the current main of datasets, you see no difference between datasets, and no throughput improvement for different gpus per model. The max DL throughput is ~17it/s
image

with this PR, we see throughout improvements from using the chunked datasets alongside model sharding.the max throughput is ~170it/s for 16gpm. 4 grid chunks seems like a sensible middle ground, given that no real model is likely to go anywhere near that fast :D
image

src/anemoi/datasets/usage/gridded/grids.py
```diff
@@ -596,11 +596,20 @@ def __getitem__(self, index: FullIndex) -> NDArray[Any]:
...
+
     @debug_indexing
     @expand_list_indexing
     def _get_tuple(self, index: TupleIndex) -> NDArray[Any]:
         """Helper method that applies masks and retrieves data from each dataset according to the specified index.
 
+        The grid index is pushed down to the underlying datasets, so only the
+        requested region is read (e.g. only the zarr chunks it intersects),
+        instead of loading full fields and slicing in memory.
+
```
MicheleCattaneo, line 612:
```suggestion
```
I don't think this belongs in a docstring. This is an implementation description that can be an in-line comment. The user does not need to know about implementation

MicheleCattaneo (changes requested):
Only a small style request from my side. Thanks for looking into this!

src/anemoi/datasets/usage/gridded/grids.py
frazane: We already use many in-line comments in the implementation below, so a summary of what makes the method special deserves to be in a docstring IMO.

MicheleCattaneo: You are right that the in-line comments are already there, no need to add anything.
I still would not (and its just a personal opinion) put this kind of description in a docstring. Up to you if you want to resolve this and leave the comment, or wait to hear other opinions

japols (approved):
LGTM, thanks!

MicheleCattaneo (approved):
Also looks good to me, thank you!

cathalobrien (approved):
great find, thanks for the contribution

PR #749, opened 2026-09-02
feat: add icosahedral resolutions 

### LINKED ISSUE #748: icosahedral resolutions
### Is your feature request related to a problem? Please describe.

We (DWD) would like to request an extension of the resolution specification to support sub-divided icosahedral grids with the pattern `rXXbYY` (with integers XX and YY). E.g. `r3b6` or `r03b06`. in principle, we would prefer upper case `R`and `B`, but if the full string has to be lower case, `r` and `b` would be acceptable.

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### Organisation

DWD

NatalieZelenka:
## Description
Closes #748

Allows datasets to be registered in the catalogue with icosahedral resolutions (`r\d\db\d\d`).

----
📚 Documentation preview 📚: https://anemoi-datasets--749.org.readthedocs.build/en/749/


MeraX (commented):
Natalie, thanks for the quick response. This looks good in principle.
I think it would just need updating the documentation here
https://github.com/ecmwf/anemoi-datasets/blob/4c3a20aa796f600f6ce7d0a824b90d172aac0a69/docs/building/naming-conventions.rst?plain=1#L76 and in the registry package.

NatalieZelenka:
Updated the docs and added a matching PR to `anemoi-registry` https://github.com/ecmwf/anemoi-registry/pull/158

MeraX (dismissed):

MeraX:
Thank you so much. Just a final thought: I would appreciate if the documentation and the Regex code would be somewhat linked to each other.

docs/building/naming-conventions.rst
```diff
@@ -40,6 +40,10 @@ whether they are gridded or tabular.
    added. The part **purpose** is especially difficult to define for
    some datasets and may be revisited.
 
+.. note::
+
+   The naming convention is implemented in the ``anemoi.datasets.create.naming`` module.
+
```
floriankrb, line 46: This is a private API and, IMHO, should not be mentioned here. Especially since this file `naming-convention.rst` should not really belong to `anemoi-datasets` and is likely to be copied verbatim to other places.
If we do want to mention implementation details that are not public API, we should be clear about it (adding, for instance, "this implementation detail is private API and may change without any warning and without creating a breaking change").

MeraX: Hi @floriankrb, what do you mean by "this is a private API"? I though the naming convention is part of an user-facing interface since it confronts every user when creating a new dataset or registering a dataset in the catalogue, which is the place where the MLPP agreed to share their datasets. As an alternative, I would recommend describing the conventions much more detailed in the documentation, as `the dataset XYZ does not follow naming convention. See here for details: https://anemoi-registry.readthedocs.io/en/latest/naming-conventions.htm` is IMHO not very helpful for a user.

My intention of the note was to have an analogy to "if you want to use the imputer, set `processors.imputer._target_: anemoi.models.preprocessing.imputer.InputImputer`." This also implicitly links the docs to the implementation and avoids duplication of all details in the docs and the risk of implementation/documentation divergence. But maybe you have an other suggestion in mind how to support the uses?

floriankrb: The naming conventions don't belong to anemoi-datasets, it should not be here, the dataset creation part of the package `anemoi-datasets` does not decide on the naming of the datasets. On the other hand, the dataset curator sets naming conventions for the datasets in their catalogue and the people creating datasets choose the names they like. We added this check of dataset name for convenience here because we didn't have (yet) an easy way to rename a dataset safely and because when a dataset is built, it is annoying for users to have the catalogue reject it.

On the other hand, we could put it in the catalogue, but then creating a dependency on the catalogue when building a dataset would make things more difficult to build datasets from HPC nodes with no internet access, or without the right credentials to access the catalogue (this may change with the future anemoi-nexus), and in any case, requiring internet access to manipulate data locally would not be an acceptable requirement for most users.

It does not belong to `anemoi-registry` either, as this package is just a tool, a generic client to talk to any anemoi catalogue and has nothing to say about the information it communicates.
 
The clean way to do it would be to put it in another package such as anemoi-naming-conventions to avoid having extra dependencies, but this has not been done (mainly because discussion about naming conventions takes time and people don't have the bandwidth to think much about it).

All this discussion is quite far from [the original issue](https://github.com/ecmwf/anemoi-datasets/issues/748) as I understand it. I believe the commit made yesterday solves the issue that "datasets with resolutions "r..b..." in their name are wrongly rejected". If not, I misunderstood the issue. Do you want to add more about this point? this is perfectly within the scope of https://github.com/ecmwf/anemoi-registry/issues/149 : changing the naming conventions.
The other points about "we should or shouldn't duplicate code, we should or should not document the internal implementation of the naming conventions checking code, anemoi-datasets and anemoi-registry and the catalogue should or should not depend on each other" are a different topic, would you like to keep/start the conversation going about this in another issue?

NatalieZelenka:
I'll close this now since the other PR was merged. I'll open a new one soon with a fix to the docs to follow up on what we're speaking about here with @MeraX

PR #753, opened 2026-09-08
fix: mismatch between shape and missing dates

b8raoult:
We found a bug impacting datasets with missing dates. The bug was introduced in version 0.5.37 (version 0.5.36 was correct).

This bug: https://github.com/ecmwf/anemoi-datasets/issues/700,  has actually nothing to do with cutout. This is a bug in the handling of missing dates at creation time, which means that the datasets created are incorrect: the data in zarr is compressed, i.e. we skip the missing dates, and all as the result the dataset is shorter. This bug was introduced version 0.5.37
 
This fix that bugs. In addition, open_dataset will fail on badly encoded datasets.

### DIFF
--- a/src/anemoi/datasets/create/creator.py
+++ b/src/anemoi/datasets/create/creator.py
@@ -33,7 +33,7 @@ from .parts import PartFilter
 
 LOG = logging.getLogger(__name__)
 
-VERSION = "0.20"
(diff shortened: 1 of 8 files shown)


reuterbal (approved):
Looks good, many thanks!

PR #754, opened 2026-09-09
feature: optional specification of dates when using multiple netcdf files as source with paterns

yoel-zerah:
## Description
anemoi-datasets currently offers the possibility to use multiple netcdf as a source: https://anemoi.readthedocs.io/projects/datasets/en/latest/building/sources/netcdf.html#netcdf
I propose to incorporate a recipe option to specify with patterns which dates can be found in netcdf files, like this:

```yaml
dates:
  start: 2020-01-01 00:00:00
  end: 2020-12-31 23:59:59
  frequency: 3h
input:
  netcdf:
    path: /path/to/my/file_{date:strftime(%Y-%m-%d)}.nc
    date_content:
      start: "{date:strftime(%Y-%m-%d)}T00:00:00"
      end: "{date:strftime(%Y-%m-%d)}T23:59:59"
      frequency: 3h
```  
This recipe says that each file contains only a day's worth of dates. In turn, anemoi-dataset should only look for those specified dates in the files.

I'm interested in reading several netcdf at once, which can be currently done with this:
```yaml
dates:
  start: 2020-01-01 00:00:00
  end: 2020-12-31 23:59:59
  frequency: 3h
input:
  netcdf:
    path: /path/to/my/file_{date:strftime(%Y-%m-%d)}.nc
```  
This is compatible with my netcdf files of daily 3h data for a year (366 files for year 2020), that are saved under a format like `file_2020_01_05.nc`.

While anemoi-datasets manages to compile and create the dataset, it is quite inefficient and throws many warnings:
```
2026-09-09 15:34:28 WARNING Could not find valid_datetime=2020-01-02T18:00:00 in TimeCoordinate[name=time,values=8,shape=(8,)] (alias of time)
```
This is because anemoi-datasets checks for the presence of each dataset date in each netcdf file, and returns a warning for each missing date (so (366-1)*366*8=1068720 warning lines in my case).

None

##  Additional notes ##
At the moment, I have only implemented a solution for netcdfs, though I guess it would work well for xarrays, or actually any source. 
In effect, the proposed change would allow to build anemoi datasets with netcdfs while having a similar file layout as gribs: https://anemoi.readthedocs.io/projects/datasets/en/latest/building/sources/grib.html#grib
I am quite bad at naming things, so I'd appreciate any proposition to make the optional recipe parameters better, and the code clearer.

### DIFF
--- a/src/anemoi/datasets/create/sources/netcdf.py
+++ b/src/anemoi/datasets/create/sources/netcdf.py
@@ -42,4 +42,5 @@ class NetCDFSource(LegacySource):
         object
             The loaded data.
         """
-        return load_many("📁", context, dates, path, *args, **kwargs)
+
+        return load_many("📁", context, dates, path, date_content=kwargs.pop("date_content", None), *args, **kwargs)
--- a/src/anemoi/datasets/create/sources/patterns.py
+++ b/src/anemoi/datasets/create/sources/patterns.py
@@ -13,6 +13,7 @@ from collections.abc import Generator
 from typing import Any
 
 from earthkit.data.utils.patterns import Pattern
+from anemoi.datasets.dates import DatesProvider
 
 
 def _expand(paths: list[str]) -> Generator[str, None, None]:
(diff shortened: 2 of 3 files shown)


NatalieZelenka:
Hello, we already do have some functionality like this: https://anemoi.readthedocs.io/projects/datasets/en/latest/howtos/create/01-grib-data.html#reading-grib-messages-from-files-that-follow-a-pattern. Is there a reason this wouldn't work for you?

In case it's helpful, here's an example  of a dataset with pattern matching and netcdf files (but beware it is an oldish recipe so not ALL the yaml structure is the same now, but I think pattern-matching should be) http://anemoi.ecmwf.int/datasets/wfde5-ea-an-oper-0001-mars-0p5-1979-2019-1h-v1?tab=recipe

yoel-zerah:
```yaml
build:
   group_by: daily
```
enabled to fix my issue without requiring new feature implementation.

PR #759, opened 2026-09-14
fix: Make odb source handle select wildcard correctly and make pivot robust to duplicates

### LINKED ISSUE #760: ODB source: pivoting fails when duplicate rows exist for an index/pivot-key combination
**Describe the bug**

`pivot_obs_df` calls `pandas.DataFrame.pivot()` directly on the input data. If the ODB data contains duplicate rows for the same index/pivot-key combination (which can happen with real-world ODB inputs, as discovered during testing), `pivot()` raises `ValueError: Index contains duplicate entries, cannot reshape`, and dataset creation fails with no easy way to recover or understand why.

**Version number**

Current `main` / this branch of `anemoi-datasets`.

**To Reproduce**

1. Create an ODB dataset with `pivot_columns`/`pivot_values` configured, using input data that contains two or more rows sharing the same combination of index columns and pivot column values.
2. Dataset creation fails with a `ValueError` from `pandas.DataFrame.pivot()` instead of completing.

**URL to sample input data**

N/A — reproducible with any dataframe input containing duplicate rows for the same index/pivot-key combination; no external sample file required.

**Expected behavior**

Pivoting should be resilient to duplicate rows: duplicates on the index/pivot-key combination should be dropped (keeping the first occurrence) before pivoting, and a warning should be logged so users have observability into when/why rows were dropped, rather than the operation failing outright.

### LINKED ISSUE #761: ODB source: `select` expressions containing `*` unintentionally disable required columns
**Describe the bug**

`odb_sql_str` disables the default/required columns (date, time, latitude, longitude, and any `required_columns`) whenever the substring `"*"` appears anywhere in the `select` expression. This means a `select` value such as `count(*)`, the use of `*` for multiplication, or any expression that merely contains a `*` as part of a larger clause, unintentionally disables the required columns — even though the user did not ask for a full wildcard select. Required columns should only be disabled when `select` is *explicitly* `"*"` (optionally with surrounding whitespace).

**Version number**

Current `main` / this branch of `anemoi-datasets`.

**To Reproduce**

1. Configure an `odb` source with `select: "count(*)"` and a flavour with `required_columns` (e.g. `obsvalue`).
2. Observe that the generated SQL statement omits the date/time/lat/lon columns and `obsvalue`, even though `select` was not intended as a full wildcard.

**URL to sample input data**

N/A — reproducible with any `select` expression containing `*` as part of a larger clause; no external sample file required.

**Expected behavior**

Required/default columns should only be disabled when `select` is explicitly `"*"` (ignoring surrounding whitespace). Any `select` expression that merely contains `*` as part of a larger clause (e.g. `count(*)`) should still include the default/required columns.

ReubenHill:
## Description
Fixes two ODB source edge cases in dataset creation:
- Makes wildcard handling in `select` robust so default required columns are only disabled when the user explicitly sets select to `"*"`.
- Makes pivoting resilient to duplicate rows by deduplicating on pivot keys before pivot and logging a warning when duplicates are dropped.

This is a bugfix.

Without these changes:
- A select expression containing `*` as part of a larger expression will unintentionally disable default required columns.
- Pivoting could fail when duplicate rows exist for the same index/pivot-key combination.

With these changes:
- Wildcard behaviour is explicit and predictable.
- Pivoting is robust for noisy ODB inputs and provides observability through warnings.

- Task: Problems found when reading some ODB data
- Issue: resolves #760 and resolves #761 


ReubenHill:
@b8raoult I can't select reviewers for this - the [contribution guide](https://anemoi.readthedocs.io/en/latest/contributing/contributing.html#code-review-process) suggests I ought to be able to but I clearly don't have the necessary permissions.

floriankrb:
This looks like a nice fix for he two issues you pointed out. Thanks!

floriankrb:
But shouldn't this logic be in anemoi-transform? @aaron-hopkinson

Issue #767, opened 2026-09-17
accumulate: no way to correct archives that encode startStep=0 on per-interval fields


mpvginde:
**Describe the bug**

The `accumulate` source cannot consume archives that stamp `startStep=0` on fields which are really *per-interval* values. There is no patch that expresses the correction, so every such field is rejected with `Field not used for any accumulation` and the build dies.

Concrete case: `class=rr`, `origin=se-al-ec`, `param=10fg`. The archive offers fields at end steps 1–6 hourly then 3-hourly (9, 12, 15, …). The field at step 9 covers `[6, 9]`, but it is encoded as `startStep=0, endStep=9`. `endStep` is correct; only `startStep` is wrong.

`FieldToInterval` reconstructs each field's window from `startStep`/`endStep`, so it places that field at `[12:00, 21:00]` while the accumulator needs `[18:00, 21:00]`. `match_interval` compares `(min, max, base)` exactly, nothing matches, and the field is rejected.

The two existing patches cannot express the fix:

- `set_start_step_to_zero` → `startStep = 0` (already 0, no change)
- `reset_24h_accumulations` → `startStep = 9 - 9 % 24 = 0` (no change)

Nor would any simple arithmetic rule: the steps are hourly to 6 and 3-hourly after, so neither `endStep - 1` nor `endStep - 3` is right across the range.

** Version number **

`0.5.41`

**To Reproduce**

```yaml
- accumulate:
    period: 6h
    covering:
      auto:
        - [0,  "0-1/1-2/2-3/3-4/4-5/5-6/6-9/9-12/12-15/15-18/18-21/21-24"]
        - [12, "0-1/1-2/2-3/3-4/4-5/5-6/6-9/9-12/12-15/15-18/18-21/21-24"]
    source:
      mars:
        class: rr
        expver: prod
        origin: se-al-ec
        stream: oper
        type: fc
        levtype: sfc
        param: [10fg]
```

For valid date 2020-12-31 00:00 the source correctly resolves the covering to `[6-9]` and `[9-12]` off base 2020-12-30 12:00, and fires the right MARS requests. The fields come back encoded `0-9` and `0-12`:

```
💬 Existing accumulators (1) :
  Accumulator(valid_date=2020-12-31 00:00:00, 6h, key={ class=rr, ..., param=10fg, ... })
    todo: SignedInterval(20201230.1800+3h ->         2100, base=20201230.1200, [6-9] )
[...]

Issue #567, opened 2026-03-10
Add `reduce` source for arbitrary temporal reductions


frazane:
### Problem

The current `accumulate` source in anemoi-datasets covers the specific case of temporally accumulated fields very well, with a sophisticated logic to determine covering intervals required to compute a target accumulation. However, there is no equivalent source for other temporal reductions, such as mean, max/min, std, etc. Users might need a way to apply other common temporal reductions declaratively in dataset recipes. Concretely, we would currently need this to compute hourly maxima of wind (i.e. wind gust) starting from 10min maxima.

### Proposed solution
Add a new `reduce` source (or generalise `accumulate`) to support arbitrary reductions. The interface should mirror `accumulate`:

```yaml
input:
  reduce:
    source: {mars: ...}
    period: 6h
    operation: mean  # or max, min, sum, std, etc.
    availability: ...
```

Importantly, this new source would reuse much of the interval-handling machinery already implemented for the `accumulate` source.

frazane:
Bump because I still think this would be a worthy addition!

elkir:
We just found we _need to use this_ feature to correctly "accumulate" wind gusts `fg10`, min and max temperature `mn2t`/`mx2t`, with the `min( )` or `max( )`.

This is because in CERRA forecasts, all the steps +1h, +2h, +3h exist, so each is accumulating only the previous hour. And for a correct 3-hourly dataset, we need to reduce them together.

frazane:
> We just found we _need to use this_ feature to correctly "accumulate" wind gusts `fg10`, min and max temperature `mn2t`/`mx2t`, with the `min( )` or `max( )`.
> 
> This is because in CERRA forecasts, all the steps +1h, +2h, +3h exist, so each is accumulating only the previous hour. And for a correct 3-hourly dataset, we need to reduce them together.

@elkir is this something on which you would like to work?

PR #768, opened 2026-09-17
feat: add a reduce source

mpvginde:
## Summary

Add a new source, `reduce`, that builds one output field per validity date by applying a single operation — `sum`, `max`, `min` or `mean` — to the archived fields covering `[valid_date - period, valid_date]`.

`accumulate` stays as it is. `reduce` reuses its machinery (covering search, grouping, metadata patching, GRIB output) but is deliberately narrower in what it is allowed to do to get a window.

## Motivation

Some parameters are not accumulations but *windowed statistics*. A wind gust field holds the largest gust observed since the last output, not a quantity accumulated over it.

Concrete case: an archive provides hourly maximum wind gust (`10fg`), the field valid at `T` covering `[T-1h, T]`. For a dataset at 6h frequency the value at `T` should be

```
gust_6h(T) = max( g[T-6h,T-5h], g[T-5h,T-4h], …, g[T-1h,T] )
```

Those are exactly the six fields `accumulate` already retrieves for a 6h window — the covering search, the grouping, the interval bookkeeping and the output encoding are all identical. Only the combination step differs, and today it is hardcoded to addition in `Accumulator.compute`:

```python
local_values = matching.sign * values.copy()
if self.values is None:
    self.values = local_values
else:
    self.values += local_values
```

Addresses #567 and #767 
## Scope

`reduce` is exclusively for quantities that can be aggregated **directly**, by applying the declared operation to fields that already tile the window.

Reconstructing a window by *differencing* archived fields (`a(6,12) = +a(0,12) − a(0,6)`) is **not** in scope — that is what `accumulate` is for. `reduce` never subtracts, and this holds for `sum` too, not only for the extrema.

| | `accumulate` | `reduce` |
|---|---|---|
| Combines with | addition only | `sum`, `max`, `min`, `mean` |
| May subtract archived fields | **yes** | no, never |
| Needs the window tiled exactly | no | **yes** |
| Works on accumulated-from-start archives | yes | no |
| Accepts instantaneous fields | n/a | no — rejected explicitly |
| Trajectory recipes | yes | not initially |

## Why the restriction is not arbitrary

Addition is invertible; `max` is not. The maximum over `[6,12]` cannot be recovered from the maxima over `[0,12]` and `[0,6]`. So a non-invertible operation is only defined when the covering is an all-positive tiling of the window: every interval positive, union exactly `[valid_date - period, valid_date]`, nothing hanging outside it.

The good news is that this needs no new search logic. Restricting the candidate set to positive intervals makes the existing Dijkstra produce exactly the required contiguous tiling, because the walk then advances monotonically from `start` to `end`. Filtering *during* the search rather than rejecting the result afterwards matters: the search minimises total length, so it can return a `+9h −3h` path even where a clean positive tiling exists, and the failure would otherwise only surface after the retrieval had been paid for.

## Declarative API

```yaml
input:
  reduce:
    period: 6h
    operation: max          # sum (default) | max | min | mean
    covering:
      auto:
        - [0,  "0-1/1-2/2-3/…"]
        - [12, "0-1/1-2/2-3/…"]
    source:
      mars: {class: od, stream: oper, type: fc, levtype: sfc, param: [10fg]}
```

`covering:`, `patch:` and `group_by:` keep the same shapes as in `accumulate`.

Supported sources: `mars`, `fdb`, `grib-index`.

## Operations

| Value | Meaning |
|---|---|
| `sum` | Default. Adds the covering fields. |
| `max` | Largest value over the window, element by element (e.g. maximum wind gust). |
| `min` | Smallest value over the window. |
| `mean` | Time-weighted average over the window. |

`mean` must weight each field by the length of the interval it covers:

```
mean = sum(length_i * values_i) / sum(length_i)
```

This is needed. An archive that is hourly out to step 6 and 3-hourly beyond it (e.g. `rr`/`se-al-ec`) can tile a window with a 1h piece and a 3h piece, and an unweighted average of the two wrong.

`mean` is well defined here precisely *because* `reduce` requires an exact tiling: the pieces are disjoint and their lengths sum to the period, so there is no double counting and no need to know what to subtract. (It would not be well defined on an overlapping or signed covering, which is a further argument for the forward-only rule above.)

Implementation note: the operation needs the interval length, which the combine step does not currently receive, so this implies a small extension — pass the contributing interval's length, plus a "finalise" hook where `mean` divides by the accumulated total. `sum`/`max`/`min` ignore both.

##Design choices that are open for discussion

**Instantaneous fields must be rejected, loudly.** This is the most likely way to misuse the source: "6-hourly mean 2t" sounds like something it should do, and it is not. A snapshot covers no interval (`startStep == endStep`) and tiles nothing.

Worse, the current code does not merely fail to support it — it silently produces a wrong answer. `FieldToInterval` reads a field with `startStep == endStep` as covering `[0, endStep]`, so an instantaneous 2t at step 6 becomes "covers `[0h, 6h]`", tiles a 6h window on its own, and is written out as `stepType=avg` over that window:

```
value=280.0  (the snapshot was 280.0)
stepType=avg window=0-6 valid=20210101/0600
```

The builder then ingests it happily as `process=average`. `reduce` must therefore refuse a field with `startStep == endStep` rather than applying that fallback, and the error should name both possible causes, because the condition is ambiguous:

- the parameter is instantaneous → wrong source for the job;
- it is an interval statistic with broken metadata → repairable with a `patch:` (see #767).

Note `accumulate` genuinely relies on the `[0, endStep]` fallback — its own GRIB1 accumulations are written as `stepType=instant` with `startStep == endStep`, and `create/gridded/result.py` has the reciprocal hack to read them back — so the stricter behaviour must be opt-in for `reduce`, not a change to the shared path.

**De-aggregation is out of scope.** If a parameter is stored accumulated from the start of the forecast, getting a window out of it would require de-accumulating first. `reduce` should detect this while resolving the covering — before any data is retrieved — and fail with an error that says so and points at `accumulate` for additive quantities. A future development could add de-aggregation.

**`covering: auto` is precipitation-shaped.** The presets in `_match_mars_config` are keyed on `(class, stream, origin)` only, not on `param`, and their step layouts were written for precipitation. A parameter such as wind gust may be archived with entirely different step ranges in the same class/stream, so `auto` is likely to hand back the wrong candidate intervals. Recipes will want an explicit interval list, and the failure should say so rather than emitting a bare "cannot find coverage".

## Some issues with the GRIB intermediate result as discovered by CLAUDE.ai
**GRIB edition 1 cannot encode an extremum, and whether that bites depends on the parameter.** (This concerns encoding the result, so it only applies to fields that passed the interval check above — an instantaneous parameter is rejected earlier, whatever the operation.) GRIB1's time range indicator (table 5) has codes for average (3) and accumulation (4) but none for maximum or minimum, so eccodes writes both extrema as `timeRangeIndicator=2` ("valid over the interval") and reads that back as a maximum whatever was written. In GRIB1 the statistic is carried by the *parameter* instead — `10fg` is a maximum, `mn2t6` a minimum, `tp` an accumulation — exposed as the param-derived key `stepTypeForConversion`, which is what `result.py` falls back on.

Round-tripping every combination through `result.py`'s resolution chain (eccodes 2.47), writing edition 1 throughout:

| edition | parameter declares | operation | recorded as | |
|---|---|---|---|---|
| 1 | `max` (`10fg`) | `max` | `maximum` | ok |
| 1 | `max` (`10fg`) | `min` | `maximum` | **silently wrong** |
| 1 | something else (`tp` declares `accum`) | `max` or `min` | *nothing* | **raises later in the build** |
| 1 | any | `sum`, `mean` | `accumulation` / `average` | ok, via trI 4 / 3 |
| 2 | any | any | correct | ok, via `typeOfStatisticalProcessing` |

So the guard cannot simply be "reject `min` on GRIB1": `max` on a non-max parameter is equally broken, just with a later and more confusing failure.

Refusing outright is also too strict, because some archives are GRIB1-only (`rr`/`se-al-ec` is), which would put a legitimate operation permanently out of reach — the maximum of hourly `tp` ("wettest hour in the window"), for instance.

**Proposal: promote the output to GRIB2 when edition 1 cannot express the operation.** The GRIB written by this source is an internal intermediate, so the edition is invisible in the resulting dataset. Verified: the statistic is then recorded correctly (`typeOfStatisticalProcessing`), the values are exact, and the MARS local definition (`class`/`stream`/`expver`/`type`) survives the conversion.

The cost is that eccodes resolves the promoted message to the parameter that genuinely matches the statistic, so the output parameter is **renamed**: a 6h minimum of `10fg` becomes `min_i10fg` ("Time-minimum 10 metre wind gust"). For combinations with no canonical GRIB2 parameter — the maximum of an accumulation such as `tp` — it becomes `unknown`. Values and statistic are still right, but a `rename:` keyed on the original name will not match, so the promotion should be logged.

Promotion should apply only where it is needed: an extremum whose parameter already declares that statistic stays in edition 1 unchanged, as do `sum` and `mean` for any parameter.

**A known gap, inherited from `accumulate`.** `sum` on a GRIB1 parameter that does not itself declare `accum` (summing `10fg`, say) goes through the historical encoding — `stepType=instant` with `step=period` — from which `result.py` records no `process` or `period` at all, because its recovery path requires `stepTypeForConversion == "accum"`. The values are correct and nothing is mislabelled; only the provenance is missing. Writing the proper `timeRangeIndicator=4` form instead would fix it for any parameter, but that would change `accumulate`'s output encoding, so it is left out of this proposal.

**`stepType` must be written explicitly, not inherited from the template.** The template handed to the writer is whichever contributing field happened to arrive last. Inheriting its `stepType` works only when that field is already a windowed statistic; on an `instant` template the output silently comes out as `stepType=instant` with the window collapsed to zero length, with only an `ECCODES WARNING` and no exception. This matters downstream: `create/gridded/result.py` derives a `process`/`period` from the output field and **raises** if it cannot recognise the statistic. With an explicit `stepType`, a 6h max resolves to `process=maximum, period=(0h, 6h)` in both editions.

Related gotcha for whoever touches `writers.py`: earthkit passes the metadata dict to `codes_set_key_vals` in insertion order, so `stepType` must be set **before** `startStep`/`endStep`. Reversing them collapses the window with only a warning. The current code gets this right by argument order alone.

(All of the above verified locally with eccodes 2.47 / earthkit-data 0.16.3.)

## Out of scope for a first iteration

- **Averaging instantaneous fields.** A "mean 2t over 6h" built from snapshots is a legitimate thing to want, but it is a different operation: it needs a covering of *sample instants* rather than a tiling of intervals, which the `SignedInterval` machinery does not express. Worth its own proposal; `reduce` should reject it clearly in the meantime.
- **Trajectory recipes** — the trajectory branch ignores `covering:`, so `reduce` has no declaration to resolve. Should raise a clear error for now.
- **De-aggregating accumulated-from-start archives.**

---

Version: observed against `0.5.41`.

----
📚 Documentation preview 📚: https://anemoi-datasets--768.org.readthedocs.build/en/768/

### DIFF
--- a/docs/building/sources/accumulate.rst
+++ b/docs/building/sources/accumulate.rst
@@ -45,6 +45,9 @@ The ``accumulate`` source requires the following parameters:
   release with a ``DeprecationWarning``.
 - **accumulation** (trajectory recipes only): one of ``from-zero`` or
   ``from-previous-step``; see `Forecast accumulations (trajectory recipes)`_.
+- **reduction** (optional, default ``sum``): how the fields covering the window
+  are combined — ``sum``, ``max`` or ``min``. Use ``max`` for windowed extrema
+  such as wind gusts; see `Reductions: sum, max and min`_.
 - **patch** (optional): Patches to apply to fields returned by the source to fix metadata issues.
   Default patching is to set ``startStep`` to ``0`` when ``startStep==endStep``.
 
@@ -188,6 +191,91 @@ For full control, provide an explicit list of ``(basetime, steps)`` pairs.
           :language: yaml
 
 These two examples are equivalent to those shown in Option 1 above.
+
+Reductions: sum, max and min
+============================
+
+By default the fields covering a window are **added** together — the classic
+accumulation for precipitation or radiation. Some parameters are instead
+*windowed extrema*: a wind gust field holds the largest gust observed since the
+last output, not a quantity accumulated over it. Adding six hourly gust fields
+would be meaningless; you want their maximum.
+
+Set ``reduction`` to choose:
+
+.. list-table::
+   :widths: 15 85
+   :header-rows: 1
+
+   * - Value
+     - Meaning
+   * - ``sum``
+     - Default. Adds the covering fields, subtracting where the covering uses a
+       reversed interval.
+   * - ``max``
+     - Largest value over the window, element by element (e.g. maximum wind gust).
+   * - ``min``
+     - Smallest value over the window.
+
+For example, a 6-hourly dataset built from hourly gust fields:
+
+.. code:: yaml
+
+   input:
+     accumulate:
+       period: 6h
+       reduction: max
+       covering:
+         auto:
+           - [0,  "0-1/1-2/2-3/3-4/4-5/5-6/6-7/7-8/8-9/9-10/10-11/11-12"]
+           - [12, "0-1/1-2/2-3/3-4/4-5/5-6/6-7/7-8/8-9/9-10/10-11/11-12"]
+       source:
+         mars:
+           class: od
+           stream: oper
+           type: fc
+           levtype: sfc
+           param: [10fg]
+
+For a valid date of 12:00 this takes the maximum of the six hourly fields
+covering 06:00 → 12:00.
+
+What ``max`` and ``min`` require
+--------------------------------
+
+Addition is invertible, so a window can be reconstructed by *subtracting*
+archived fields: ``a(6,12) = +a(0,12) - a(0,6)``. This is what makes
+``accumulated-from-start`` archives usable, and what ``accumulation: from-zero``
+relies on in trajectory recipes.
+
+Extrema are **not** invertible. The maximum over ``[6,12]`` cannot be recovered
+from the maxima over ``[0,12]`` and ``[0,6]``. Consequently ``max`` and ``min``
+only work when the archive provides intervals that **tile the window exactly**,
+which in practice means per-step values:
+
+- An archive storing per-step values (ERA5-style, ``accumulated-from-previous-step``,
+  or an explicit ``0-1/1-2/...`` interval list) works.
+- An archive storing values from the start of the forecast
+  (``accumulated-from-start``) does **not**, and the source raises rather than
+  producing a wrong answer.
+- In trajectory recipes, ``accumulation: from-previous-step`` works and
+  ``accumulation: from-zero`` is rejected.
+
+.. warning::
+
+   The ``covering: {auto: ...}`` presets describe how **precipitation** is laid
+   out for a given class/stream. A parameter such as wind gust may be archived
+   with completely different step ranges in the very same stream, so give an
+   explicit interval list rather than relying on ``auto``. If no forward-only
+   covering can be found, the source fails with an error naming the window.
+
+.. note::
+
+   GRIB edition 1 has no way to encode a minimum: it shares
+   ``timeRangeIndicator=2`` with maximum, so the field would be read back as a
+   maximum. ``reduction: min`` therefore raises an error when the source
+   delivers GRIB1; ``max`` is encoded correctly in both editions. NaNs
+   propagate, as they do for ``sum``.
 
 Controlling the fields regrouped within accumulation
 ====================================================
--- a/src/anemoi/datasets/create/sources/accumulate/accumulator.py
+++ b/src/anemoi/datasets/create/sources/accumulate/accumulator.py
@@ -17,6 +17,8 @@ from numpy.typing import NDArray
 
 from anemoi.datasets.create.intervals import SignedInterval
 
+from .reductions import Reduction
+from .reductions import reduction_factory
 from .writers import write_accumulated_field_with_valid_time
 from .writers import write_accumulated_forecast_field
 
@@ -34,6 +36,7 @@ class Accumulator:
         key: dict[str, Any],
         coverage,
         basetime: datetime.datetime | None = None,
+        reduction: Reduction | str | None = None,
     ):
         # The accumulator only accumulates fields and does not know about the rest
         # Accumulator object for a given param/member/valid_date
@@ -42,6 +45,7 @@ class Accumulator:
         self.period = period
         self.key = key
         self.basetime = basetime
+        self.reduction = reduction_factory(reduction)
 
         self.coverage = coverage
 
@@ -54,21 +58,25 @@ class Accumulator:
         """Check whether the accumulation is complete (all intervals have been processed)"""
         return not self.todo
 
-    def compute(self, values: NDArray, interval: SignedInterval) -> None:
-        """Perform accumulation with the values array on this interval and record the operation.
+    def compute(self, values: NDArray, interval: SignedInterval) -> bool:
+        """Fold the values array into this accumulator if it needs this interval.
+
         Note: values have been extracted from field before the call to `compute`,
         so values are read from field only once.
 
-        Parameters:
+        Parameters
         ----------
-        field: Any
-            An earthkit-data-like field
         values: NDArray
-            Values from the field, will be added to the held values array
+            Values from the field, combined into the held values array according
+            to ``self.reduction``.  The array is shared with the other
+            accumulators this field contributes to and is never mutated here.
+        interval: SignedInterval
+            The interval the field covers.
 
         Return
         ------
-        None
+        bool
+            True if the interval was needed and consumed, False otherwise.
         """
 
         def match_interval(interval: SignedInterval, lst: list[SignedInterval]) -> bool:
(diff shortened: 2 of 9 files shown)


src/anemoi/datasets/create/sources/accumulate/operations.py
frazane, line 1: This module is also used by the `reduce` subpackage, perhaps we could move it one level up the hierarchy?

docs/building/sources/accumulate.rst
```diff
@@ -189,6 +190,75 @@ For full control, provide an explicit list of ``(basetime, steps)`` pairs.
 
 These two examples are equivalent to those shown in Option 1 above.
 
+Fixing mis-encoded step metadata
```
frazane, line 193: How common is this mis-encoding? If it is very rare, or only for a specific dataset, I would not document it here, but only concisely in the code: the few people that need it will know about it.

docs/building/sources/reduce.rst
```diff
@@ -0,0 +1,265 @@
...
+   mean = sum(length_i * values_i) / sum(length_i)
+
+This matters whenever the covering mixes interval lengths. An archive that is
+hourly out to step 6 and 3-hourly beyond it can tile a window with a 1h piece
+and a 3h piece; a plain unweighted average of the two would be wrong. Because
+``reduce`` requires the intervals to tile the window exactly, the weights sum to
+the requested period and nothing is double counted. With a uniform covering the
+result is the ordinary arithmetic mean.
+
+Instantaneous fields are not supported
```
frazane, line 122: What if I have instantaneous 10min temperature and I want the daily max/min? It is a fairly common use case.

src/anemoi/datasets/create/sources/reduce/source.py
```diff
@@ -0,0 +1,133 @@
...
+
+from ..accumulate.field_to_interval import FieldToInterval
+from ..accumulate.operations import operation_factory
+from ..accumulate.source import AccumulateSource
+
+LOG = logging.getLogger(__name__)
+
+
+@source_registry.register("reduce")
+class ReduceSource(AccumulateSource):
```
frazane, line 52: I don't think inheritance is the right approach here. Note it does not really work the other way around either: accumulate reconstructs a window by subtracting archived fields, so it is not a fold over a tiling, and neither source is a subtype of the other. I think these two classes should be completely separate, sharing modules rather than a base class.

src/anemoi/datasets/create/sources/accumulate/accumulator.py
frazane, line 28: If I understand correctly, this class performs both reductions and accumulations even with subtractions? If yes,  I have doubts this is the right way. I would separate the two concerns: direct, simple reductions should be one thing, and the more complex accumulations with the djkstra approach should be a separate code path only used in the AccumulateSource.

src/anemoi/datasets/create/sources/accumulate/source.py
```diff
@@ -93,6 +101,8 @@ def __init__(
         self.period = frequency_to_timedelta(period)
         self.covering = covering
         self.accumulation = accumulation
+        # accumulate always adds; `reduce` is the source that varies the operation.
+        self.operation = Sum()
```
frazane, line 105: This is not true though, no? The AccumulateSource can also subtract to get accumulations. I would not use an `Operation` here. Also relates to my comment on the `Accumulator`.

frazane (commented):
Hi @mpvginde, I had a high level view of the changes and I am not convinced by the overall design choices. The most problematic aspects for me are the class inheritance and mixing the "direct" reduction logic with the more complex accumulation logic that also handles subtractions. Those should be separate things.

Of course I agree with the intent of maximizing code reuse and have modularity, but I would propose an alternative approach: keep `accumulate` and `reduce` as two separate sources (no inheritance), and achieve what we want by extracting what they share into new modules that they then both use. For example the covering search or interval generation, but I am sure there are other things.

As for the rest, GRIB metadata in particular, I don't fully understand yet but it seems like you've done a thorough analysis and picked up on some important edge cases, thanks!

src/anemoi/datasets/create/sources/accumulate/accumulator.py
mpvginde: Before, this class was basically always summing:

The class gets a list of fields (each of them with a specific sign) from the `covering` machinery and sums them with the specific sign (if the sing is negative, its a decumulation. positive is an accumulation). 

So the dijkstra approach lives somewhere else.

The changes introduced here generalizes the operation performed on the fields. So the functionality of the class goes from: 
I get a bunch of fields that I should sum 
to 
I get a bunch of fields and I should apply a user-defined operation on them.

floriankrb:
Please keep in mind that this PR changes code that has been refactored a lot in the branch feat/hindcasts (accumulations) https://github.com/ecmwf/anemoi-datasets/pull/713

mpvginde:
@floriankrb do I read it correct that #713 also implements (temporal) reductions with min, max and avg?

